PR #1, opened 2020-08-19
many: add spread testing using GCE backend for ubuntu-core-initramfs

anonymouse64:
This sets up the project to be testable via spread, both locally via nested QEMU
VM's, as well as in the cloud with GCE nested VM's.

There are some TODO's still around simplifying this even more and getting rid of
more cruft from snapd's setup, but this is sufficient right now to get a nested
UC20 VM with our version of the initramfs in it for testing.

To run the tests locally via qemu, first install qemu, then get spread via 

```
$ sudo apt update && sudo apt install -y qemu-kvm autopkgtest
$ curl https://niemeyer.s3.amazonaws.com/spread-amd64.tar.gz | tar -xz -C $GOPATH/bin
```

Then make your qemu spread images dir with

```
$ mkdir -p ~/.spread/qemu
$ cd ~/.spread/qemu
```

and make a suitable spread image with:

```
$ autopkgtest-buildvm-ubuntu-cloud -r focal
$ mv autopkgtest-focal-amd64.img ubuntu-20.04-64.img
```

then finally you can run spread tests like so:

```
$ spread qemu-nested
2020-08-19 13:31:03 Project content is packed for delivery (9.00MB).
2020-08-19 13:31:03 Sequence of jobs produced with -seed=1597861863
2020-08-19 13:31:03 If killed, discard servers with: spread -reuse-pid=86651 -discard
2020-08-19 13:31:03 Allocating qemu-nested:ubuntu-20.04-64...
2020-08-19 13:31:03 Serial and monitor for qemu-nested:ubuntu-20.04-64 available at ports 59414 and 59514.
2020-08-19 13:31:03 Waiting for qemu-nested:ubuntu-20.04-64 to make SSH available...
2020-08-19 13:31:04 Allocated qemu-nested:ubuntu-20.04-64.
2020-08-19 13:31:04 Connecting to qemu-nested:ubuntu-20.04-64...
2020-08-19 13:31:22 Connected to qemu-nested:ubuntu-20.04-64 at localhost:59314.
2020-08-19 13:31:22 Sending project content to qemu-nested:ubuntu-20.04-64...
2020-08-19 13:31:23 Preparing qemu-nested:ubuntu-20.04-64 (qemu-nested:ubuntu-20.04-64)...
2020-08-19 13:45:03 Executing qemu-nested:ubuntu-20.04-64:tests/spread/basic:noenc (qemu-nested:ubuntu-20.04-64) (1/4)...
2020-08-19 13:53:49 Restoring qemu-nested:ubuntu-20.04-64:tests/spread/basic:noenc (qemu-nested:ubuntu-20.04-64)...
2020-08-19 13:53:49 Executing qemu-nested:ubuntu-20.04-64:tests/spread/basic:fullenc (qemu-nested:ubuntu-20.04-64) (2/4)...
2020-08-19 14:04:44 Restoring qemu-nested:ubuntu-20.04-64:tests/spread/basic:fullenc (qemu-nested:ubuntu-20.04-64)...
2020-08-19 14:04:45 Executing qemu-nested:ubuntu-20.04-64:tests/spread/basic:onlysecboot (qemu-nested:ubuntu-20.04-64) (3/4)...
2020-08-19 14:13:36 Restoring qemu-nested:ubuntu-20.04-64:tests/spread/basic:onlysecboot (qemu-nested:ubuntu-20.04-64)...
2020-08-19 14:13:36 Executing qemu-nested:ubuntu-20.04-64:tests/spread/basic:onlytpm (qemu-nested:ubuntu-20.04-64) (4/4)...
2020-08-19 14:24:44 Restoring qemu-nested:ubuntu-20.04-64:tests/spread/basic:onlytpm (qemu-nested:ubuntu-20.04-64)...
2020-08-19 14:24:45 Discarding qemu-nested:ubuntu-20.04-64...
2020-08-19 14:24:45 Successful tasks: 4
2020-08-19 14:24:45 Aborted tasks: 0
```

This takes approximately an hour for me on my well spec'd desktop machine, largely because we can't use nested KVM accelerated VM's, so the L2 nested VM runs significantly slower.

To run the tests via GCE, you would just run

```
$ spread google-nested
```

but to do so you need GCE credentials, see https://github.com/snapcore/spread#google

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -1,0 +1,122 @@
+project: core-initrd
+
+environment:
+  SETUPDIR: /home/core-initrd
+  PROJECT_PATH: $SETUPDIR
+  PATH: $PATH:$PROJECT_PATH/tests/bin
+  TESTSLIB: $PROJECT_PATH/tests/lib
+  TESTSTOOLS: $PROJECT_PATH/tests/lib/tools
+  # TODO: are these vars needed still?
+  LANG: "C.UTF-8"
+  LANGUAGE: "en"
+
+backends:
+  google-nested:
+    type: google
+    key: '$(HOST: echo "$SPREAD_GOOGLE_KEY")'
+    location: computeengine/us-east1-b
+    plan: n1-standard-2
+    halt-timeout: 2h
+    systems:
+      - ubuntu-20.04-64:
+          image: ubuntu-2004-64-virt-enabled
+          storage: 20G
+
+  qemu-nested:
+    type: qemu
+    # TODO:UC20: uc20 needs 2G or grub will not loopback the kernel snap
+    #            we use nested VM's, so use 4G for the host until we can boot
+    #            uc20 with less RAM
+    memory: 4G
+    systems:
+      - ubuntu-20.04-64:
+          username: ubuntu
+          password: ubuntu
+
+path: /home/core-initrd
+
+exclude:
+  - .git
+  - "*.o"
+  - "*.a"
+  - "*.snap"
+  - "*.img"
+
+prepare: |
+  # NOTE: This part of the code needs to be in spread.yaml as it runs before
+  # the rest of the source code (including the tests/lib directory) is
+  # around. The purpose of this code is to fix some connectivity issues setup 
+  # spread functions to be more usable outside of the explicit task.yaml 
+  # sections.
+
+  # apt update is hanging on security.ubuntu.com with IPv6, prefer IPv4 over IPv6
+  cat <<EOF > gai.conf
+  precedence  ::1/128       50
+  precedence  ::/0          40
+  precedence  2002::/16     30
+  precedence ::/96          20
+  precedence ::ffff:0:0/96 100
+  EOF
+  if ! mv gai.conf /etc/gai.conf; then
+      echo "/etc/gai.conf is not writable, ubuntu-core system? apt update won't be affected in that case"
+      rm -f gai.conf
+  fi
+
+  # Take the MATCH and REBOOT functions from spread and allow our shell
+  # scripts to use them as shell commands. The replacements are real
+  # executables in tests/lib/bin (which is on PATH) but they source
+  # spread-funcs.sh written here, base on the definitions provided by SPREAD.
+  # This ensures that 1) spread functions define the code 2) both MATCH and
+  # REBOOT are executables and not functions, and can be called from any
+  # context.
+  type MATCH | tail -n +2 > "$TESTSLIB"/spread-funcs.sh
+  unset MATCH
+  type REBOOT | tail -n +2 >> "$TESTSLIB"/spread-funcs.sh
+  unset REBOOT
+
+  # prepare common uc20 image setup by repacking snaps, etc
+  "$TESTSLIB"/prepare-uc20.sh
+
+restore-each: |
+  # delete the nested VM image after each task finishes so we don't use too much
+  # disk space
+  . "$TESTSLIB/nested.sh"
+  cleanup_nested_core_vm
+
+# it takes a while to build everything while preparing the project
+warn-timeout: 20m
+# but it shouldn't take _too_ long...
+kill-timeout: 30m
+
+suites:
+  tests/spread/:
+    summary: Main spread tests
+    environment:
+      # since we are only testing uc20 here, disable kvm unless the env
+      # specifically knows kvm works and sets this to true
+      # kvm specifically does not work on GCE with focal L1 VM and focal (UC20)
+      # L2 VM's because the L2 VM intermittently and randomly hangs and GCE's
+      # hypervisor will restart the VM, there is not currently a known
+      # workaround
+      ENABLE_KVM: '$(HOST: echo "${SPREAD_ENABLE_KVM:-false}")'
+
+      # variants of various system setup below, turning on/off secure boot and
+      # enabling/disabling tpm, this ensures that all tests we write will be run
+      # against all variants
+
+      # TODO: are there other variants/knobs the initramfs we want to support
+      # testing?
+
+      # full encryption has TPM and secure boot
+      ENABLE_TPM/fullenc: true
+      ENABLE_SECURE_BOOT/fullenc: true
+      # only secboot does not have a TPM
+      ENABLE_TPM/onlysecboot: false
+      ENABLE_SECURE_BOOT/onlysecboot: true
+      # only tpm does not have secure boot
+      ENABLE_TPM/onlytpm: true
+      ENABLE_SECURE_BOOT/onlytpm: false
+      # no enc does not have secure boot or tpm
+      ENABLE_TPM/noenc: false
+      ENABLE_SECURE_BOOT/noenc: false
+# vim:ts=4:sw=4:et
--- a/tests/lib/nested.sh
+++ b/tests/lib/nested.sh
@@ -1,0 +1,169 @@
+#!/bin/bash
+
+WORK_DIR="${WORK_DIR:-/tmp/work-dir}"
+SSH_PORT=${SSH_PORT:-8022}
+MON_PORT=${MON_PORT:-8888}
+IMAGE_FILE="${WORK_DIR}/ubuntu-core-20.img"
+
+execute_remote(){
+    sshpass -p ubuntu ssh -p "$SSH_PORT" -o ConnectTimeout=10 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ubuntu@localhost "$*"
+}
+
+wait_for_service() {
+    local service_name="$1"
+    local state="${2:-active}"
+    for i in $(seq 300); do
+        if systemctl show -p ActiveState "$service_name" | grep -q "ActiveState=$state"; then
+            return
+        fi
+        # show debug output every 1min
+        if [ "$i" -gt 0 ] && [ $(( i % 60 )) = 0 ]; then
+            systemctl status "$service_name" || true;
+        fi
+        sleep 1;
+    done
+
+    echo "service $service_name did not start"
+    exit 1
+}
+
+wait_for_ssh(){
+    retry=400
+    wait=1
+    while ! execute_remote true; do
+        retry=$(( retry - 1 ))
+        if [ $retry -le 0 ]; then
+            echo "Timed out waiting for ssh. Aborting!"
+            return 1
+        fi
+        sleep "$wait"
+    done
+}
+
+cleanup_nested_core_vm(){
+    # stop the VM if it is running
+    systemctl stop nested-vm-*
+
+    # remove the swtpm
+    # TODO: we could just remove/reset the swtpm instead of removing the snap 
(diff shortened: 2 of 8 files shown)


tests/lib/prepare-uc20.sh
```diff
@@ -0,0 +1,281 @@
...
+# extract the kernel snap, including extracting the initrd from the kernel.efi
+kerneldir=/tmp/kernel-workdir
+"$TESTSLIB/repack-kernel.sh" extract upstream-pc-kernel.snap $kerneldir
+
+# extract snap-bootstrap from the snapd snap and put into the kernel work dir
+cp /tmp/snap-bootstrap-from-snapd-snap $kerneldir/skeleton/main/usr/lib/snapd/snap-bootstrap
+
+# copy the skeleton from our installed ubuntu-core-initramfs into the initrd 
+# skeleton for the kernel snap
+cp -ar /usr/lib/ubuntu-core-initramfs/main/* "$kerneldir/skeleton/main"
```
anonymouse64, line 189: this in effect means that we are using the same version of kernel modules from the snap that were already there, the same kernel, a new snap-bootstrap from edge channel of snapd, and a new version of the ubuntu-core-initramfs layout from spread/the PR under test.

I seem to remember there being some thinking that what we were doing in snapd was not great and that we should be using some other location/source for the kernel modules or something, but perhaps I am just misremembering

tests/lib/repack-kernel.sh
```diff
@@ -0,0 +1,233 @@
+#!/bin/bash
```
anonymouse64, line 1: ideally $SOMEDAY this whole script becomes a command on the ubuntu-core-initramfs snap

tests/spread/basic/task.yaml
```diff
@@ -0,0 +1,36 @@
...
+  # Now formally wait for seeding
+  execute_remote sudo snap wait system seed.loaded
+
+  # We should be in run mode
+  execute_remote cat /proc/cmdline | MATCH snapd_recovery_mode=run
+
+  # We should have a modeenv
+  execute_remote test -f /var/lib/snapd/modeenv
+
+  # TODO: what else qualifies a "successful boot"?
```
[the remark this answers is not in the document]
anonymouse64: I should mention that we have tests for recover mode via spread in snapd, but the setup is complicated, but we do not have any tests for install mode, so setting up tests for install mode is very hard and setting up tests for recover mode is just "kinda hard".

tests/lib/prepare-uc20.sh
```diff
@@ -0,0 +1,281 @@
...
+
+if getent group systemd-journal >/dev/null; then
+    usermod -G systemd-journal -a test
+    id test | MATCH systemd-journal
+fi
+
+echo 'test ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
+
+# get the model
+# TODO: should this be a variant to test multiple grades i.e. signed and secured?
```
anonymouse64, line 44: looking at this again, it occurs to me that we can't specify unasserted snaps for signed or secured models, which means we couldn't inject users into the VM's for those grades, so probably this TODO should just go away and we are stuck with dangerous only grades.

restricting ourselves to just dangerous also means we could just use cloud-init instead of the whole special user setup too :thinking:

xnox:
> This takes approximately an hour for me on my well spec'd desktop machine, largely because we can't use nested KVM accelerated VM's, so the L2 nested VM runs significantly slower.

Why are nested VMs used locally? Can spread not launch the VM under test from my local machine directly? Can environment be fixed to allow that?

spread.yaml
```diff
@@ -0,0 +1,122 @@
...
+  - "*.img"
+
+prepare: |
+  # NOTE: This part of the code needs to be in spread.yaml as it runs before
+  # the rest of the source code (including the tests/lib directory) is
+  # around. The purpose of this code is to fix some connectivity issues setup 
+  # spread functions to be more usable outside of the explicit task.yaml 
+  # sections.
+
+  # apt update is hanging on security.ubuntu.com with IPv6, prefer IPv4 over IPv6
```
xnox, line 52: huh, no way. Is there a bug about this? Are you sure that nested VMs have correct ipv6 passthrough setup?

spread.yaml
```diff
@@ -0,0 +1,122 @@
...
+      ENABLE_SECURE_BOOT/fullenc: true
+      # only secboot does not have a TPM
+      ENABLE_TPM/onlysecboot: false
+      ENABLE_SECURE_BOOT/onlysecboot: true
+      # only tpm does not have secure boot
+      ENABLE_TPM/onlytpm: true
+      ENABLE_SECURE_BOOT/onlytpm: false
+      # no enc does not have secure boot or tpm
+      ENABLE_TPM/noenc: false
+      ENABLE_SECURE_BOOT/noenc: false
```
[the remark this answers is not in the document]
xnox: I don't like using the name "secboot" in the variant name, especially in the one that doesn't refer to what we call secboot (measured and secured boot with FDE).

Also 3 out of 4 variants are effectively the same. Because without either secureboot or without tpm or without both, we just do plain boot.

Can't we reduce this to just two variants, always with a tpm, and one variant with secureboot on, second variant with secureboot off?

Or is there something special about testting secureboot-off with tpm or secureboot-on without tpm? I don't think there is any different code paths for those, are there?

tests/lib/nested.sh
```diff
@@ -0,0 +1,169 @@
...
+    PARAM_SERIAL="-serial file:${WORK_DIR}/serial.log"
+    PARAM_TPM=""
+
+    ATTR_KVM=""
+    if [ "$ENABLE_KVM" = "true" ]; then
+        ATTR_KVM=",accel=kvm"
+        # CPU can be defined just when kvm is enabled
+        PARAM_CPU="-cpu host"
+        # Increase the number of cpus used once the issue related to kvm and ovmf is fixed
+        # https://bugs.launchpad.net/ubuntu/+source/kvm/+bug/1872803
```
xnox, line 91: Above was fixed in june, so can we try again with more CPU cores nested in Google cloud?

tests/lib/prepare-uc20.sh
```diff
@@ -0,0 +1,282 @@
...
+    cp ../*.deb "$SETUPDIR"
+)
+
+# install ubuntu-core-initramfs here so the repack-kernel.sh uses the version of
+# ubuntu-core-initramfs we built here
+apt install -yqq "$SETUPDIR"/ubuntu-core-initramfs*.deb
+
+# now download snap dependencies 
+# TODO:UC20: when should some of these things start tracking stable ?
+snap download pc-kernel --channel=20/edge --basename=upstream-pc-kernel
```
xnox, line 65: ooooh, i didn't know about `--basename` arg.

tests/lib/prepare-uc20.sh
xnox: there was, and when we tried that, we found bugs =) which were not fixed yet.

It would be nicer to execute ubuntu-core-initramfs to create and sign kernel.efi in one go, for the kernel/modules extracted from the snap. But that's not possible yet.

Also not sure if during this test we want to use existing snap-bootstrap from inside the kernel-snap, the one from snapd edge, or the one that is "current" / stable / vendored inside ubuntu-core-initramfs.

tests/spread/basic/task.yaml
```diff
@@ -0,0 +1,36 @@
...
+  # We should be in run mode
+  execute_remote cat /proc/cmdline | MATCH snapd_recovery_mode=run
+
+  # We should have a modeenv
+  execute_remote test -f /var/lib/snapd/modeenv
+
+  # TODO: what else qualifies a "successful boot"? 
+  # * no failed systemd units?
+  # * all expected partitions mounted?
+  # * systemd not in degraded mode?
```
xnox, line 36: LUKS has been setup for ubuntu-data, when expected.

TPM measurements were completed.

snapd has sealed things.

xnox (approved):
There are a few todo's that we should revisit, but this is mostly ok and helps automating some of our testing.

Should I have already expected a check executed on this PR? or does it need to be merged first before adding github actions to it?

anonymouse64:
> > This takes approximately an hour for me on my well spec'd desktop machine, largely because we can't use nested KVM accelerated VM's, so the L2 nested VM runs significantly slower.
> 
> Why are nested VMs used locally? Can spread not launch the VM under test from my local machine directly? Can environment be fixed to allow that?

That would require changes to spread, it can be done, but then you have to deal with the fact that someone could be running spread on this system from their xenial machine (or even Fedora!) and not have new QEMU, and not have all the other dependencies, etc. and makes the output less predictable and potentially fails with random problems due to these dependencies. With this setup, all you need is spread + some qemu and it will always work

tests/spread/basic/task.yaml
anonymouse64: Sure, I will leave it to you to write such tests in followups, I just left this as a clear entry-point for foundations to take over testing and write whatever tests are appropriate

spread.yaml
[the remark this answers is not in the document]
anonymouse64: Sure if you want to eliminate 2 of the variants to just have secure boot w/ TPM (and thus also full disk encryption), as well as the plain boot without any of this that's fine. I included them because there is code in snap-bootstrap that operates slightly differently depending on if we have just secure boot or just a TPM, etc. but we already have tests for this different behavior in snapd, so that setup doesn't need to be replicated here.

tests/lib/prepare-uc20.sh
anonymouse64: yes I also am not sure what the right thing to do is, you probably saw in the PR, but I actually changed it slightly after opening the PR to not stage snap-bootstrap from the snapd snap into the kernel snap's initramfs, but actually use the version of snap-bootstrap that comes from snapd from the snappy-dev/image PPA so that this test tests exactly what ubuntu-core-initramfs would do if it was built today.

that being said it would still be kinda nice if we have testing somewhere here like a nightly or something that ran with ubuntu-core-initramfs from main here, but also staged snap-bootstrap from edge snapd snap into it too to make sure that continues to work and doesn't regress as edge changes

anonymouse64:
> Should I have already expected a check executed on this PR? or does it need to be merged first before adding github actions to it?

No, I spoke with Gustavo, he preferred to merge the spread setup first then get the keys setup, I will work on doing that now with him since we have merged this however.

tests/lib/nested.sh
anonymouse64: sure go for it, but every time we try to use faster / better VM's in GCE it doesn't work for us in weird ways, so YMMV

spread.yaml
anonymouse64: well this was originally added to snapd's spread.yaml as of https://github.com/snapcore/snapd/pull/3418, so perhaps things are better now, I will leave it to you to look into removing this, if it works for you to remove it, then let us know and we'd be happy to remove the same from our spread.yaml :-)

PR #5, opened 2020-09-11
.travis.yml: basic spread testing via travis

anonymouse64:
WIP as we setup the required keys, etc. for this to actually work


xnox:
travis? not github actions? or will we have both?

anonymouse64:
it was easier to use travis actually, so just travis

PR #8, opened 2020-09-30
the-modeenv: bind mount /boot/efi -> /run/mnt/ubuntu-seed

mvo5:
This commit ensures the ESP efi dir is available for firmware
update tooling. This fixes LP: 1892392

### DIFF
--- a/factory/usr/lib/the-modeenv
+++ b/factory/usr/lib/the-modeenv
@@ -12,6 +12,8 @@ if grep -q snapd_recovery_mode=run /proc/cmdline; then
     # TODO:UC20: support other bootloaders too
     if [ -f /run/mnt/ubuntu-boot/EFI/ubuntu/grub.cfg ]; then
         echo '/run/mnt/ubuntu-boot/EFI/ubuntu /boot/grub none bind 0 0' >> /run/image.fstab
+        # ensure ESP efi dir is available for fwupdate (LP: 1892392)
+        echo '/run/mnt/ubuntu-seed/ /boot/efi none bind 0 0' >> /run/image.fstab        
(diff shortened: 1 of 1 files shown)


xnox:
Hi, this has been discussed in-depth on Mattermost, that doing so, will not enable fwupd snap to work; and will make it easier to break FDE.

Are we sure we still want to expose ESP like that? Especially since it will not fix https://bugs.launchpad.net/snapd/+bug/1892392 until https://github.com/fwupd/fwupd/issues/2513 is fixed.

xnox:
This would be ok to do today, but only for non-FDE cases. Or for example, fwupd to ensure that it doesn't try to apply updates when FDE is present, as without resealing support it will lock the users out / prevent booting into run mode.

superm1:
>This would be ok to do today, but only for non-FDE cases. Or for example, fwupd to ensure that it doesn't try to apply updates when FDE is present, as without resealing support it will lock the users out / prevent booting into run mode.

Is it sealing to PCR0 or PCR2?  If not, then UEFI firmware updates shouldn't cause problems I would expect.

xnox:
> > This would be ok to do today, but only for non-FDE cases. Or for example, fwupd to ensure that it doesn't try to apply updates when FDE is present, as without resealing support it will lock the users out / prevent booting into run mode.
> 
> Is it sealing to PCR0 or PCR2? If not, then UEFI firmware updates shouldn't cause problems I would expect.

PCR2, PCR4, PCR7, PCR12 are all used to seal the secret against.

xnox:
There has been progress to make this work with fwupdate upstream, thus this PR will help with that effort. But this alone will not fix it all for FDE sealed boots.

xnox (approved):

PR #11, opened 2020-11-11
the-tool: log stdout, stderr to journal and console

anonymouse64:
This is important for field debugging where we want to be able to see why
"the-tool" failed.


bboozzoo (approved):
Pretty please

xnox (approved):

xnox:
spread tests didn't work though =/

anonymouse64:
Ah the tests failed because they take a long time to run, and produce no output, let me see if I can adjust the "no output" limit on travis for this, if not, then I can also make spread be more noisy too

PR #12, opened 2020-11-16
spread.yaml: drop warn-timeout to 5 minutes

anonymouse64:
Hopefully now spread will provide output every 5 minutes while waiting for things to finish such that travis doesn't think that the job is infinitely stuck.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,4 @@ install:
   - popd
 
 script:
-  - env
   - "/tmp/spread google-nested:"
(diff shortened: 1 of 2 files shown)


anonymouse64:
mmm
```
Still running (50 of 60): /tmp/spread google-nested:
The job exceeded the maximum time limit for jobs, and has been terminated.
```
perhaps it's just too slow on GCE? it doesn't take that long to run on GCE when I launch it manually here for some reason...

anonymouse64:
I fear we need to move from travis to github actions instead, so I'm going to close this I will try to work on getting this repo setup with GitHub actions soonish

PR #20, opened 2021-05-19
repack-kernel.sh: when repackging kernel only the main feature is ava…

xnox:
repack-kernel.sh: when repackging kernel only the main feature is available


xnox:
@mvo5 @anonymouse64 @sil2100 hopefully this will fix the new spread tests.

I guess the top level binary should be improved to not autoadd features that are not available from a skeleton.

anonymouse64:
I'm curious if this works, it didn't when I tried it locally running spread tests but who knows

xnox:
@anonymouse64 it now worked!

anonymouse64:
nice!!! :tada:

PR #21, opened 2021-05-19
Add sbat, arm64, armhf, riscv64 support

xnox:


xnox:
@anonymouse64 so the github actions tested the tip of my proposed branch, rather than the automatic merge. Is that expected?

Cause this branch originally was forked from before main branch got fixed. Then main branch was fixed up, yet 66f778b still failed. So i had to merge in main to pass the tests in 842ff56.

Somehow expected the CI to take the PR /merge head for testing.

anonymouse64:
@xnox I don't know why that happened, but I think reading up on github actions/checkout@v2, I think that the default behavior of v2 is to use the merge commit, so maybe we should try switching to v2 ?

PR #22, opened 2021-05-19
Xnox/cloudimg rootfs draft 1

xnox:


xnox:
see https://github.com/snapcore/core-initrd/pull/27 instead

PR #23, opened 2021-05-21
Xnox/add on push

xnox:
add more testing schedules.


xnox:
I don't know if this is the right github actions syntax to add weekly testing & testing the main branch after it gets merged/pushed.

anonymouse64 (approved):
this looks correct to me to do a weekly build on sunday of tests, as well as to do builds on the main branch on every push there

PR #24, opened 2021-05-21
.github/workflows/tests.yml: use checkout v2

anonymouse64:
This PR does not include #20, so let's see if using checkout v2 causes us to use the merge commit for tests


anonymouse64:
This seems to have passed and seems to have run with the generated merge commit, so I think we should merge this and then we can avoid issues like what happened in https://github.com/snapcore/core-initrd/pull/21

PR #25, opened 2021-06-04
Add cmdline bin/ubuntu-core-initramfs

xnox:


xnox:
CI fails from time to time, but this is not changing anything in the initrd. I somehow suspect CI is flakey. Let's merge this and see what the CI says fro the push result.

PR #26, opened 2021-06-07
Xnox/revert sbat

xnox:
CI seems to fail after this change got merged.


xnox:
@chrisccoulson somehow my backports from upstream of sbat stuff fail CI.

I will revert sbat support for now, which i guess is not urgent given we need a fixed up shim first.

PR #30, opened 2021-06-15
Xnox/simple sbat

xnox:


xnox:
Rebuilding kernel snap with this change in place, boots UC20 with secureboot correctly, with tpm, and FDE.

xnox:
This works with shim-signed from impish-proposed, with secureboot on, on classic too.

PR #32, opened 2021-07-27
Changes to support arm64 in the systemd EFI stub

alfonsosanchezbeato:
First draft to support aarch64 in EFI stub. Main changes are

- The way the kernel is started in arm64 is different to x86 one: instead of using boot parameters, it takes a device tree that needs to be modified to specify the initrd location. Also, we are not calling it directly but via the kernel EFI stub, that takes command line from the EFI parameters so we do not need to touch them in systemd stub.
- objcopy does not support PE/COFF for aarch64, so to create kernel.efi we use llvm-objcopy instead. Even in that case, llvm-objcopy seems to be buggy and some fields in the PE headers need to be modified from the script.

I think that changes in sources need to go to a patch, but for the moment I am proposing directly to ease review.


anonymouse64:
(I didn't review, but I trust @alfonsosanchezbeato so I let the CI run to see what happens)

anonymouse64:
Ah actually, due to how the spread setup is for core-initrd, the CI cannot run on a branch from a fork like Alfonso has done here, not sure how folks want to handle this...

bin/ubuntu-core-aarch64-efi
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/python3
```
stulluk, line 1: Please add Canonical header and the purpose

stulluk (changes requested):

vendor/systemd/src/boot/efi/stdclib.c
```diff
@@ -0,0 +1,78 @@
...
+    const char *ret = 0;
+    do {
+        if (*s == (char)c)
+            ret = s;
+    } while(*s++);
+    return (char *)ret;
+}
+
+uintptr_t __stack_chk_guard = 0xdeadbeefa55a857;
+
```
stulluk, line 75: Please explain the purpose of this function

vendor/systemd/src/boot/efi/linux-aarch64.c
```diff
@@ -0,0 +1,157 @@
...
+#include <efi.h>
+#include <efilib.h>
+#include <libfdt.h>
+
+#include "linux.h"
+#include "linux-aarch64.h"
+
+#ifndef EFI_DTB_TABLE_GUID
+#define EFI_DTB_TABLE_GUID \
+    { 0xb1b621d5, 0xf19c, 0x41a5, {0x83, 0x0b, 0xd9, 0x15, 0x2c, 0x69, 0xaa, 0xe0} }
```
stulluk, line 12: Where does those magic values come from ?

alfonsosanchezbeato:
Overriden by https://github.com/snapcore/core-initrd/pull/35

PR #34, opened 2021-08-10
features: enabled server feature on x86_64 and any generic kernels.

xnox:
Currently server feature is only enabled by default on x86_64
architecture. Also enable server feature for generic kernels, as they
are likely to boot in Qemu too.

BugLink: https://bugs.launchpad.net/bugs/1938530
Signed-off-by: Dimitri John Ledkov <dimitri.ledkov@canonical.com>

### DIFF
--- a/bin/ubuntu-core-initramfs
+++ b/bin/ubuntu-core-initramfs
@@ -238,7 +238,11 @@ def main():
 
     args = parser.parse_args()
     if args.subcmd == "create-initrd" and not args.features:
+        # Set arch specific default set of features
         args.features = features.split()
+        # For generic kernels, enable more server drivers too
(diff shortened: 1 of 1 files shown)


xnox:
Needs manual test by building pc-kernel test snap for arm64 & amd64.

alfonsosanchezbeato:
I have tested this now for both amd64 & arm64. With the patch, the generated arm64 initramfs includes `virtio_blk.ko` and drive is detected when using qemu option `-drive if=virtio,...`

alfonsosanchezbeato (approved):
LGTM, @xnox I think this can be unblocked now.

valentindavid (approved):
LGTM.

PR #35, opened 2021-08-17
Changes to support arm64 in the systemd EFI stub

alfonsosanchezbeato:
First draft to support aarch64 in EFI stub. Main changes are

* The way the kernel is started in arm64 is different to x86 one: instead of using boot parameters, it takes a device tree that needs to be modified to specify the initrd location. Also, we are not calling it directly but via the kernel EFI stub, that takes command line from the EFI parameters so we do not need to touch them in systemd stub.
* objcopy does not support PE/COFF for aarch64, so to create kernel.efi we use llvm-objcopy instead. Even in that case, llvm-objcopy seems to be buggy and some fields in the PE headers need to be modified from the script.

I think that changes in sources need to go to a patch, but for the moment I am proposing directly to ease review.

### DIFF
--- a/bin/ubuntu-core-aarch64-efi
+++ b/bin/ubuntu-core-aarch64-efi
@@ -1,0 +1,179 @@
+#!/usr/bin/python3
+# Copyright 2021 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import os
+import shutil
+import subprocess
+import tempfile
+
+# Sizes of different parts of the PE+ file
+sz_signature = 4
+sz_coff_fhead = 20
+sz_opt_head = 112
+sz_sect_head = 40
+
+# File offset to the PE header pointer
+off_pe_pointer = 0x3c
+# Offsets inside file header
+off_fhead_NumberOfSections = 2
+off_fhead_NumberOfSymbols = 12
+# Offsets inside optional headers
+off_opth_SizeOfInitializedData = 8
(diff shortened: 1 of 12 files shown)


xnox (changes requested):
Vendored systemd changes are normally done as patches under vendor/systemd/debian/patches => such that it is clear which things are patched here, on top of ubuntu patchset. However the way it is done here is also fine for now, as it aides the review.

The systemd changes must be submitted upstream, at least as a pull request, such that they can be reviewed there, and eventually merged. Also the arm64 systemd patches should also probably go into at least Ubuntu Devel series.

@alfonsosanchezbeato Have you submitted the systemd portion of changes upstream, or do you want me to rebase and do that?

alfonsosanchezbeato:
> Vendored systemd changes are normally done as patches under vendor/systemd/debian/patches => such that it is clear which things are patched here, on top of ubuntu patchset. However the way it is done here is also fine for now, as it aides the review.
> 
> The systemd changes must be submitted upstream, at least as a pull request, such that they can be reviewed there, and eventually merged. Also the arm64 systemd patches should also probably go into at least Ubuntu Devel series.
>

Sure, I agree with all this. The main purpose of the PR is to have a first round of reviews, and then, after addressing the comments, propose to upstream.
 
> @alfonsosanchezbeato Have you submitted the systemd portion of changes upstream, or do you want me to rebase and do that?

No, I prefer to get your review before trying that. I'm fine with proposing upstream myself, although of course some help from you side is appreciated :)

xnox:
> > @alfonsosanchezbeato Have you submitted the systemd portion of changes upstream, or do you want me to rebase and do that?
> 
> No, I prefer to get your review before trying that. I'm fine with proposing upstream myself, although of course some help from you side is appreciated :)

The systemd patches look ok. Please cherrypick them onto systemd tip, drop the "systemd: " prefix from those commits and propose them upstream as three PRs imho:

 * "add initial trace" should be a standalone PR, cause it's not related to anything else.

 * "fix linux_exec prototype" should be a standalone PR, without including anything else.

 * The rest of the arm64 patches should be standalone PR, but can include the fix linux_exec prototype commit.

Cause you don't want "add initial trace" stall prototype fix, or adding arm64 support.

alfonsosanchezbeato:
I have created now two PRs for systemd:
https://github.com/systemd/systemd/pull/20481 (small bit, already merged)
https://github.com/systemd/systemd/pull/20601
By some comments in the second one, it looks like they prefer silent boots when all is fine, so I won't propose one for adding an initial trace.

xnox:
Upstream has merged alternative implementation of arm booting. If it is sufficient it should be integrated into systemd jammy and into either systemd focal or core-initrd as extra patches.

PR #36, opened 2021-08-17
initrd-switch-root.service.d: provide a bit more context

mvo5:
This is the result of my goose-chase why `systemd-analyze time` does not
break out the time spend in initrd. It turns out that for this to work
systemd needs to re-exec statefully. But because of issues with systemd
and/or the way we use it this can't be done right now. So to ensures
that this knowledge is preserved I added some more comments to the
relevant override.

Thanks to Dimitri for (patiently) explaining all this.


xnox (approved):

stulluk:
As per our discussion with mvo, he mentions "systemd-analyze time" **after** the pivot-root here. Because there is no "systemd-analyze" binary inside the initrd at all.

stulluk (approved):
Pretty nice explanation, thanks.

PR #38, opened 2021-08-17
debian: add git to build-depends

mvo5:
git is currently needed to build the core-initrd so it should be
part of the build-depends. Without that e.g. `spread qemu-nested:`
fails early.

Thanks for Sertac for spotting this.


alfonsosanchezbeato (approved):
LGTM, thanks!

stulluk (approved):

PR #39, opened 2021-08-18
spread: increase timeouts and fixes

mvo5:
There are currently a bunch of spread failures. It looks like they
might be related to the fact that the test take a long time. This
is also evident from seeing some "green" runs. This commit increases
the timeouts to have more consistent results.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -83,6 +83,10 @@ restore-each: |
   . "$TESTSLIB/nested.sh"
   cleanup_nested_core_vm
 
+debug-each: |
+  . "$TESTSLIB/nested.sh"
+  print_nested_status
+
 # it takes a while to build everything while preparing the project
 warn-timeout: 40m
 # but it shouldn't take _too_ long...
--- a/spread.yaml
+++ b/spread.yaml
@@ -84,9 +84,9 @@ restore-each: |
   cleanup_nested_core_vm
 
 # it takes a while to build everything while preparing the project
-warn-timeout: 20m
+warn-timeout: 40m
 # but it shouldn't take _too_ long...
-kill-timeout: 30m
+kill-timeout: 50m
 
 suites:
   tests/spread/:
--- a/tests/lib/nested.sh
+++ b/tests/lib/nested.sh
@@ -28,9 +28,15 @@ wait_for_service() {
 }
 
 wait_for_ssh(){
+    local service_name="$1"
     retry=800
     wait=1
     while ! execute_remote true; do
+        if ! systemctl show -p ActiveState "$service_name" | grep -q "ActiveState=active"; then
+            echo "Service no longer active"
+            return 1
+        fi
+
         retry=$(( retry - 1 ))
         if [ $retry -le 0 ]; then
             echo "Timed out waiting for ssh. Aborting!"
@@ -169,7 +175,7 @@ EOF
     fi
 
     # Wait until ssh is ready
-    if ! wait_for_ssh; then
+    if ! wait_for_ssh "${SVC_NAME}"; then
         return 1
     fi
 }
--- a/tests/lib/nested.sh
+++ b/tests/lib/nested.sh
@@ -29,7 +29,7 @@ wait_for_service() {
 
 wait_for_ssh(){
     local service_name="$1"
-    retry=400
+    retry=800
     wait=1
     while ! execute_remote true; do
         if ! systemctl show -p ActiveState "$service_name" | grep -q "ActiveState=active"; then
@@ -129,7 +129,7 @@ start_nested_core_vm_unit(){
     if [ "${ENABLE_TPM}" = "true" ]; then
         if ! snap list swtpm-mvo > /dev/null; then
             snap install swtpm-mvo --beta
-            retry=10
+            retry=60
             while ! test -S /var/snap/swtpm-mvo/current/swtpm-sock; do
                 retry=$(( retry - 1 ))
                 if [ $retry -le 0 ]; then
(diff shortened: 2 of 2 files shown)


mardy (approved):
Yes please :-)

mvo5:
This worked once now, I trigger another build to see if it was just chance or if it really helps with the issue.

anonymouse64 (approved):
lgtm, sadly

stulluk:
If I am not mistaken, this still fails in my "very slow" local system...

https://pastebin.ubuntu.com/p/4ttwPHT7Fz/

spread.yaml
```diff
@@ -84,9 +84,9 @@ restore-each: |
   cleanup_nested_core_vm
 
 # it takes a while to build everything while preparing the project
-warn-timeout: 20m
+warn-timeout: 40m
 # but it shouldn't take _too_ long...
-kill-timeout: 30m
+kill-timeout: 50m
```
anonymouse64, line 89: maybe try 70m as the kill-timeout and 50m as the warn-timeout?

mvo5:
Fwiw, after adding the debug output it appears that the swtpm socket was not ready. So qemu was briefly in "active" state and then failed a couple of seocnds later but the code in "wait_ssh" is not smart enough to notice that nested-vm-* is no longer active. I pushed fixes for the above issues. Maybe there is more though, let's see.

tests/lib/nested.sh
```diff
@@ -117,6 +129,10 @@ start_nested_core_vm_unit(){
     if [ "${ENABLE_TPM}" = "true" ]; then
         if ! snap list swtpm-mvo > /dev/null; then
             snap install swtpm-mvo --beta
+            retry=10
```
anonymouse64, line 132: as terrible and snapd specific as the nested.sh file in snapd is, it would probably serve us well to make that sharable with this repo, since the fixes you have here are also implemented in nested.sh in snapd now, and we have separate ways of handling this error, and I'm sure there are more errors.

but anyways that is for another time I suppose, this fix looks good to me

tests/lib/nested.sh
```diff
@@ -28,9 +28,15 @@ wait_for_service() {
 }
 
 wait_for_ssh(){
-    retry=400
+    local service_name="$1"
+    retry=800
     wait=1
     while ! execute_remote true; do
+        if ! systemctl show -p ActiveState "$service_name" | grep -q "ActiveState=active"; then
```
anonymouse64, line 35: will this race with the previous command `systemctl start`? I.e. maybe the ActiveState will be "activating" or something?

anonymouse64 (approved):
lgtm, good catch with the socket. 

As I mention below, it would be cool if we figured out how to share code between snapd and this repo for i.e. nested.sh

Also a question about a new possible race between systemctl start and systemctl show given the various states that ActiveState can have (i.e. `activating`). I don't think there's a race but I am always forgetting all of the systemd interactions about what state things can be in at what times after which commands

tests/lib/nested.sh
```diff
@@ -117,6 +129,10 @@ start_nested_core_vm_unit(){
     if [ "${ENABLE_TPM}" = "true" ]; then
         if ! snap list swtpm-mvo > /dev/null; then
             snap install swtpm-mvo --beta
+            retry=10
+            while ! test -S /var/snap/swtpm-mvo/current/swtpm-sock; do
+                sleep 1
```
mvo5, line 134: Sorry, I'm a Muppet - this loop of course should decrease retry and eventually stop instead of looping forever (spread will kill it eventually of course but still a bit silly). :+100: on the idea to share code from snapd, just the fact that there is no "retry" helper is a major hindrance. Oh well - one step at a time :)

anonymouse64: Apparently /me is more of a Muppet for not noticing /o\

mvo5:
One more successful run, will restart again.

mvo5:
Second successful run, one more and I call the issue resolved :) (still needs addressing of the points that were raised in the code review.

stulluk:
@mvo5 thank you very much. I confirm it finally works in my low spec machine :

https://pastebin.ubuntu.com/p/6dWn7sGfQW/

Note: testing takes about 2 hours in my system.

Approving.

stulluk (approved):

alfonsosanchezbeato (approved):
LGTM, thanks!

mvo5:
Thanks, marking blocked until I manage to address the feedback (hopefully quick) - THANKS for all the reviews.

tests/lib/nested.sh
xnox: `systemctl start` is synchronous by default and waits for the unit to start. The asynchronous version that doesn't wait requires `--no-block` flag which is not use there. So this shouldn't be racy.  Unless of course the svc-script is racy and is not executing/started when it first spawns things. So this looks ok to me.

tests/lib/nested.sh
mvo5: Yeah, I asked Sergio to think about how we can share code between projects. I think you are spot-on here!

tests/lib/nested.sh
mvo5: I fixed this now

mvo5:
This is ready now (I addressed the issues from the review) and this can be merged when it's green.

mardy (approved):

xnox:
Note I had to retry this once to pass. Not sure if others were clicking retry as well or not. But hopefully things should get better with this in place.

mvo5:
Thanks! Let's keep an eye on it and see if there is any flakiness left.

PR #40, opened 2021-08-20
Add support for the systemd-bootchart service

alfonsosanchezbeato:
This service will be disabled by default. To run it, change the kernel command line adding `rd.systemd.wants=systemd-bootchart.service`.


debian/control
```diff
@@ -59,7 +59,8 @@ Build-Depends: debhelper-compat (= 12), dh-python, python3:any, dracut-core, qui
                libgcc-s1,
                squashfs-tools,
                snapd (>= 2.50+20.04),
-               golang-go, indent, libapparmor-dev, libcap-dev, libfuse-dev, libglib2.0-dev, liblzma-dev, liblzo2-dev, libseccomp-dev, libudev-dev, openssh-client, pkg-config, python3, python3-docutils, python3-markdown, squashfs-tools, tzdata, udev, xfslibs-dev
+               systemd-bootchart,
```
xnox, line 62: the `golang-go...` line is a carbon copy of the build-deps that snapd declares, because we are temporarily building snap-bootstrap fork, because snapd has not yet reviewed/merged changes that core-initrd relies on.

Thus it would be nice to keep that whole line unchanged for now, even though it has duplicate build-depends (squashfs-tools) declared.

factory/usr/lib/systemd/system/systemd-bootchart.service
```diff
@@ -0,0 +1,12 @@
...
+Description=Boot Process Profiler
+Documentation=man:systemd-bootchart.service(1) man:bootchart.conf(5)
+DefaultDependencies=no
+
+[Service]
+ExecStartPre=/usr/sbin/mkdir -p /run/log
+ExecStart=/lib/systemd/systemd-bootchart -r
+KillSignal=SIGHUP
+
+[Install]
```
xnox, line 11: if this is disabled by default, it shouldn't have any Install section, should it?
Or is it somehow needed?

factory/etc/systemd/bootchart.conf
```diff
@@ -0,0 +1,8 @@
+[Bootchart]
+Samples=36000
+Frequency=20
+Relative=yes
+Filter=no
+# Memory usage produces a bad overlay in the svg
```
xnox, line 6: Sadness, is this something that needs to be fixed upstream? or like make it generate memory usage separately somewhere?

stulluk (approved):
LGTM, thanks. I am still unsure about the documentation part of this functionality. It could be nice to have a single line comment in the readme.md, such as "to enable bootchart's to be created in ubuntu-save partition, just add rd.systemd... to your kernel commandline"

factory/usr/lib/systemd/system/systemd-bootchart.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Boot Process Profiler
+Documentation=man:systemd-bootchart.service(1) man:bootchart.conf(5)
+DefaultDependencies=no
+
+[Service]
+ExecStartPre=/usr/sbin/mkdir -p /run/log
+ExecStart=/lib/systemd/systemd-bootchart -r
+KillSignal=SIGHUP
```
anonymouse64, line 9: why is this here? I don't see it in the normal systemd-bootchart service on classic:

```
$ systemctl cat systemd-bootchart
# /lib/systemd/system/systemd-bootchart.service
#  This file is part of systemd.
#
#  systemd is free software; you can redistribute it and/or modify it
#  under the terms of the GNU Lesser General Public License as published by
#  the Free Software Foundation; either version 2.1 of the License, or
#  (at your option) any later version.

# Note: it's usually a better idea to run systemd-bootchart via the
# init= kernel command line switch. See the man page for details.

[Unit]
Description=Boot Process Profiler
Documentation=man:systemd-bootchart.service(1) man:bootchart.conf(5)
DefaultDependencies=no

[Service]
ExecStart=/lib/systemd/systemd-bootchart -r

[Install]
WantedBy=sysinit.target
```

factory/etc/systemd/bootchart.conf
```diff
@@ -0,0 +1,8 @@
+[Bootchart]
+Samples=36000
+Frequency=20
```
anonymouse64, line 3: so this should run for 30 minutes? that seems excessive for the initrd, no? The man page doesn't actually specify what units the Frequency is in, so I'm assuming it's Hertz, so 36000 samples / ( 20 samples / second) == 1800 seconds == 30 minutes

factory/usr/lib/systemd/system/systemd-bootchart.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Boot Process Profiler
+Documentation=man:systemd-bootchart.service(1) man:bootchart.conf(5)
+DefaultDependencies=no
+
+[Service]
+ExecStartPre=/usr/sbin/mkdir -p /run/log
```
anonymouse64, line 7: nitpick, but instead of doing this, I would rather see the config file change the output to somewhere that we don't need to mkdir -p via `Output=` but either way I suppose

anonymouse64 (commented):
can we have some sort of extension to the spread tests to check that the service runs when this is enabled? also I +1 @stulluk's comment that we should document how to enable this too

factory/usr/lib/systemd/system/systemd-bootchart.service
alfonsosanchezbeato: Right, not really needed. I've removed the install section.

stulluk:
> can we have some sort of extension to the spread tests to check that the service runs when this is enabled? also I +1 @stulluk's comment that we should document how to enable this too

@anonymouse64 +1 for considering a spread test.
In my comment about documentation, I said "I am unsure" for a reason. IMHO, this functionality should be part of our official images and we should let end-users to collect bootcharts or let them create bootcharts for us to investigate bootspeed issues.

To achieve this, IMHO, since snapd can modify kernel cmdline (considering secure boot/ without breaking signature  -- grub.cfg /EFI vars/ uboot environment), I wish there could be some way to have a functionality in snapd like "snap enable bootchart" and then Ucore creates bootcharts automatically for you. And when you collect enough bootcharts (may be after 3 reboot cycles) user can do: ``` snap disable bootchart``` and now you have 6 svg files which is pretty good way to analyze. But snapd doesn't have this feature yet, so it could be better to explain in the readme.md for now. How do you think ?

factory/usr/lib/systemd/system/systemd-bootchart.service
alfonsosanchezbeato: If killed by SIGTERM, `systemd-bootchart` does not generate the bootchart (note that the bootchart file is written right when the daemon exits). SIGHUP is handled appropriately to make sure the file is actually written. This is arguably and upstream bug, imo SIGTERM and SIGNINT should be handled in the same way as SIGHUP. But, for the moment we can specify `KillSignal` to avoid applying patches to `systemd-bootchart`.

factory/etc/systemd/bootchart.conf
alfonsosanchezbeato: 30 minutes is the maximum time. Actually, the daemon will be killed much earlier when systemd switches root, using SIGHUP for that as we specify in the unit. I've just put a big number to make sure we capture all data in extreme cases.

factory/usr/lib/systemd/system/systemd-bootchart.service
alfonsosanchezbeato: Not much is in `/run` when the initrd is running, so there are not that many options... we could directly store the chart there, but I think it is nice to have things in a subfolder so things are ordered. If you feel we really should avoid the mkdir, I would put `Output=/run` in config.

anonymouse64:
@stulluk I think that sort of feature is something it would be great to have at runtime but that's a fair amount of work. If you want you can file a launchpad bug for snapd to track enabling that sort of feature that would be appreciated, but yes in meantime something in README.md or HACKING.md is sufficient I think

alfonsosanchezbeato:
I agree that this need a spread test, it is a matter of repackaging the gadget snap with custom kernel command line. Note though that ideally we want https://github.com/snapcore/core20/pull/109 merged first, as the bootchart service in the core20 snap will move around the files generated from the initramfs to a different location, which will affect the test here. Anyway, I guess that I can give a go to the test, assuming it will need to be modified in little time.

factory/usr/lib/systemd/system/systemd-bootchart.service
anonymouse64: thanks for the explanation

factory/etc/systemd/bootchart.conf
anonymouse64: Right, I missed that the bootchart will be stopped when we switch root with SIGHUP. This is fine then.

factory/usr/lib/systemd/system/systemd-bootchart.service
anonymouse64: /run/log is fine I suppose in that case

anonymouse64:
> Note though that ideally we want snapcore/core20#109 merged first

Agreed I think landing that PR into core20 edge first will make writing a spread test here easier

factory/etc/systemd/bootchart.conf
```diff
@@ -0,0 +1,8 @@
+[Bootchart]
```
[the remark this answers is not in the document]
alfonsosanchezbeato: Done now

factory/etc/systemd/bootchart.conf
```diff
@@ -0,0 +1,8 @@
...
+Samples=36000
```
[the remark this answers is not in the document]
alfonsosanchezbeato: I have changed now the path for the chart for the core snap, so we don't need this additional change.

xnox (approved):

xnox:
Forked snapd-bootstrap fails to build at the moment, so that needs to be fixed up separately. I think govendor sync is not used correctly to checkout "old" secboot during build. As snapd has not yet been ported to a newer one.

PR #42, opened 2021-09-02
doc: add instructions on enabling bootchart

alfonsosanchezbeato:


stulluk (approved):
LGTM, thanks. However, I believe we need a separate section to describe how to modify kernel cmdline on x86 and arm respectively.

anonymouse64 (approved):
lgtm, thanks for this documentation

PR #43, opened 2021-09-02
debian: pin to commit when building snap-bootstrap

alfonsosanchezbeato:
The branch it is based on is WIP, so better fix it to something we
know is working, for the moment.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -226,7 +226,7 @@ override_dh_clean:
 ifneq ($(CI),true)
 	# only needed for the "--feature main cloudimg-rootfs", not used for the
 	# default initrd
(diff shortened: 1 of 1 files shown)


xnox (approved):

xnox:
That is the 9th of June snapshot as used in the latest PPA build from https://github.com/snapcore/snapd/pull/10267

stulluk (approved):
LGTM, but we need to create an issue for this part. We should not rely on personal forks IMHO.

xnox:
> LGTM, but we need to create an issue for this part. We should not rely on personal forks IMHO.

this is being tracked in the project that said thing relies on. In Jira, it is CPC-967 item.

PR #45, opened 2021-09-06
doc: add building.md

stulluk:
Still WIP...

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,158 @@
+# Building initrd for Ubuntu Core
+
+## Prequisities
+
+- An ubuntu desktop machine (20.04 amd64 recommended) or a VM
+- Install LXD and auto-configure it
+  ```sudo snap install lxd && lxd init --auto```
+  - Configure lxd to support storage device (a directory) mount in R/W mode
+- git
+- 2Gbyte free space in your disk
+
+## Preparation 
+
+- Fork this repository
+- Install git: 
+    ```
+    sudo apt install git
+    ```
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -0,0 +1,158 @@
...
+Attention, the package has been created in the current directory,
+not in ".." as indicated by the message above!
+root@myfocal:/src#
+```
+
+- Now we can install **build dependencies**:
+```
+apt install ./ubuntu-core-initramfs-build-deps_49_amd64.deb
+```
+This will install a ton of package. Please be patient.
```
xnox, line 87: `apt build-dep ./`

BUILDING.md
```diff
@@ -0,0 +1,158 @@
+# Building initrd for Ubuntu Core
```
xnox, line 1: This is ambigious.

First half of the document describes how to rebuild core-initrd itself. When most developers do not need to do this. They can instead simply install `ubuntu-core-initramfs` prebuilt deb package from the snappy-dev/image ppa.

Second half of the document talks about creating new intird, but not including it into kernel snap. which is more of how to use.

People who need to do first or second things are not usually the same person. Thus it would be nice to have a doc that starts with simply installing ubuntu-core-initramfs deb and using it. And separate one, to rebuild ubuntu-core-initramfs from scratch if it needs any changes.

Most people do not need to rebuild ubuntu-core-initramfs deb at all, they simply install it prebuilt.

alfonsosanchezbeato:
Closing as most of this is in HACKING.md and this is now a bit outdated.

PR #46, opened 2021-10-01
Start bootchart earlier

alfonsosanchezbeato:
Previously, we started systemd-bootchart by adding
rd.systemd.wants=systemd-bootchart.service to the kernel command
line. This has as a side effect that the service is added to the
initrd default target and is started later than desired. Instead,
enable the service by default and use a new kernel command line
parameter (core.bootchart) as condition to start it. We also need to
make sure that the service is stopped properly to the bootchart is
written to the /run folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,9 @@ TODO: Write documentation for how to integrate initrd.img to your (custom or can
 
 # Bootchart
 
-It is possible to enable bootcharts by adding
-`rd.systemd.wants=systemd-bootchart.service` to the kernel command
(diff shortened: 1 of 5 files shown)


alfonsosanchezbeato:
See also https://github.com/snapcore/core20/pull/115

mvo5 (approved):
Very nice find

anonymouse64 (approved):
lgtm

stulluk (approved):
LGTM

PR #53, opened 2021-10-27
vendor/systemd: Use a submodule instead of files in the repository

valentindavid:
We are still using the same version, that is 245.4-4ubuntu3.6 from focal. But now we can easily update to newer version of the package.


xnox:
But git subtree is already in use for this.... and one can pull systemd packaging branch to apply all the fixes.

Longer term, the goal is to land all the custom patches upstream & into Ubuntu archive and just use systemd from the Archive, such that a separate build of systemd is not needed for core-initrd.

xnox (changes requested):
subtree is already in use for this and works well enough; ideally vendored systemd should be dropped altogether with any core-initrd specific patches being upstreamed / SRUed such that no separate build of systemd is needed at all with vendor/systemd getting dropped altogether.

I don't see how above long term goals are facilitated with a switch to submodule.

PR #55, opened 2021-11-05
refactoring: a better name for snap-bootstrap

stulluk:
propose a better name for the service which
runs snap-bootstrap binary


stulluk:
wondering if there are some test mechanisms that depends on this "the-tool" naming, which would break if we make this change

anonymouse64 (commented):
We at least have code inside snapd tests that deals with "the-tool", but this can be adjusted. Probably worth trying to apply this patchset from inside snapd tests to see what happens

xnox:
Although we call snap-bootstrap, we actually care about mounts. Meaning it is the initramfs-mounts we care about. Not sure if we want a more descriptive name or not. I.e. snap-bootstrap-initramfs-mounts or even just snap-initramfs-mounts.

Also i would probably call it -wrapper rather than -starter, as it is more common.

Also i would have hoped that by now we woudn't need a -wrapper anymore and we just have ExecStart=snap-bootstrap initramfs-mounts, but oh well. So much to do, not enough time.

xnox:
Closing in favor of https://github.com/snapcore/core-initrd/pull/76/files

PR #58, opened 2021-11-19
Persist device mapper devices in udev database

valentindavid:
55-dm.rules will disable cold plugged device mapper block devices.  In
order for it not to disable them we need to keep the state from initrd
to main boot. That is we need to mark the devices as `persist_db` so
that initrd-udevadm-cleanup-db.service does not remove it.

When a device mapper device gets disabled, then all systemd mounts
with BindsTo to the device will get unmounted, as well as all bind
mounts that depend on it. Which causes a catastrophic failure of
Ubuntu Core.

As a work-around for this issue we have used a stateless reexecution
of systemd. This was making systemd forget about encrypted mount units
and thus did not trigger unmount when devices were taken down.

Now that we mark the device mapper devices as `persist_db`, this
work-around is not needed anymore.

Tested on:
 - qemu x86_64 without secure boot
 - qemu x86_64 with secure boot
 - intel nuc with secure boot
 - raspberry pi 4

The mounts in `/proc/mounts` are the same as before the fix, except
that now `systemd-mount --list` properly lists device mapper mounts.

### DIFF
--- a/factory/usr/lib/udev/rules.d/56-persist-device-mapper.rules
+++ b/factory/usr/lib/udev/rules.d/56-persist-device-mapper.rules
@@ -1,0 +1,8 @@
+# 55-dm.rules will disable in the main boot, all device mapper block
+# devices that are cold plugged. This rule file states that to keep a
+# dm device that was initialized in initrd, we are required to keep
+# its database. To prevent initrd-udevadm-cleanup-db.service from
+# removing the device from the database, we have to make all device
+# mapper block device as "db_persist".
+
+SUBSYSTEM=="block", KERNEL=="dm-[0-9]*", ACTION=="add|change", OPTIONS+="db_persist"


valentindavid:
There is some more precision of in the 55-dm.rules file:
https://github.com/lvmteam/lvm2/blob/f7730406254f023216d0727c82d46de82c124c92/udev/10-dm.rules.in#L74

stulluk (approved):
LGTM, thanks. @valentindavid could you please also describe why we are removing core-override.conf ( i.e. to be able to get systemd-analyze plot to work correctly etc ). IMHO, we definitely need to get opinions of @xnox @anonymouse64 and @mvo5 here.

BTW, with these changes, I could be able to see svg plot from my rpi4b4g succesfully.

alfonsosanchezbeato (approved):
LGTM, awesome finding! FTR https://www.spinics.net/lists/lvm/msg25180.html helped me to understand the problem, would it be possible to add it to the commit description (or maybe some other reference if you have something with a better explanation)?

xnox:
This is very interesting, thank you for digging into this. And yes it all makes sense. However, I am now perplexed as to how come this option is not already set in any udev rules that dmsetup / lvm2 / dracut ships. Because surely that is the most sensible thing to do in the initrd, before pivoting to root filesystem.

CC: @mwhudson I do wonder if we should be doing the same on the cdrom device in casper, and if that is the route cause of our troubles. Do we also cleanup udev db between initramfs-tools pivot-root?

factory/usr/lib/systemd/system/initrd-switch-root.service.d/core-override.conf
```diff
@@ -1,18 +0,0 @@
...
-# systemd from doing a stateful re-exec (there is no other way to
-# prevent this).
-#
-# Using "/sbin/init" here directly also ensures that the kernel commandline
-# "init=" parameter is ignored as it would allow to bypass our protections.
-#
-# XXX: we should fix this eventually (this may involve upstream work)
-# because this also prevents "systemd-analyze time" from displaying
-# the time spend in the initrd.
-ExecStart=/bin/systemctl --no-block switch-root /sysroot /sbin/init
```
xnox, line 18: Removing this file, will result in serialized re-exec from initrd to rootfs.  But it will also make `init=` argument on the command line work as well. which I find a bit scary. I hope that snapd prevents/prohibits sealing against customized `init=` arguments cause that would expose encryption keys in the root shell without any authentication.

@mvo5 @pedronis @chrisccoulson

anonymouse64 (commented):
nice, this makes a lot of sense, thank you for getting to the bottom of this!

Before we merge this PR, could you look into applying this patch to the initrd in snapd here: https://github.com/snapcore/snapd/blob/master/tests/lib/prepare.sh#L569-L593 

(essentially just add a line or two which deletes the override file and adds this udev rule to the unpacked initrd)

and then open a PR over there so we can see all the tests in snapd run with this patch? if things are all happy there I think this patch would then be as well tested as it possibly could be, I'm just nervous that we have missing functionality tests in this repo that are tested inside snapd spread tests and want to see those tests working first.

factory/usr/lib/systemd/system/initrd-switch-root.service.d/core-override.conf
anonymouse64: yes you cannot change kernel command line in a system with full disk encryption unless you control the gadget snap and can upload / install a new gadget snap with a new cmdline.full or cmdline.extra file, in which case we re-seal against the new kernel command line. So just booting with a init= ... whatever will not expose anything since the encrypted partition won't be able to be unlocked.

xnox:
> nice, this makes a lot of sense, thank you for getting to the bottom of this!
> 
> Before we merge this PR, could you look into applying this patch to the initrd in snapd here: https://github.com/snapcore/snapd/blob/master/tests/lib/prepare.sh#L569-L593
> 
> (essentially just add a line or two which deletes the override file and adds this udev rule to the unpacked initrd)
> 
> and then open a PR over there so we can see all the tests in snapd run with this patch? if things are all happy there I think this patch would then be as well tested as it possibly could be, I'm just nervous that we have missing functionality tests in this repo that are tested inside snapd spread tests and want to see those tests working first.

yeah, i had similar thinking. I wonder if this should be uploaded into jammy only, and be built into jammy/uc22 kernels and test it there.

although, jammy/uc22 kernel snaps would be using production signing key too.

anonymouse64:
@xnox hence just opening a PR against snapd to test with snakeoil keys in the spread tests :-)

valentindavid:
> However, I am now perplexed as to how come this option is not already set in any udev rules that dmsetup / lvm2 / dracut ships

dracut has the same rule. It should only be in an initrd context that is why it is probably not in lvm2.

xnox:
> > However, I am now perplexed as to how come this option is not already set in any udev rules that dmsetup / lvm2 / dracut ships
> 
> dracut has the same rule. It should only be in an initrd context that is why it is probably not in lvm2.

oooh, nice. I failed to spot it, which one is it? And does it have any more conditions? I am concerned on how this will work with multipath and raid.

valentindavid:
> > > However, I am now perplexed as to how come this option is not already set in any udev rules that dmsetup / lvm2 / dracut ships
> > 
> > 
> > dracut has the same rule. It should only be in an initrd context that is why it is probably not in lvm2.
> 
> oooh, nice. I failed to spot it, which one is it? And does it have any more conditions? I am concerned on how this will work with multipath and raid.

https://github.com/dracutdevs/dracut/blob/master/modules.d/90dm/11-dm.rules

xnox:
@anonymouse64 i wonder to split this PR in two. One to add udev rule => retest snapd with that.

And then separately to enable the serialized re-exec by dropping the override file.

anonymouse64:
@xnox that sounds reasonable to me

@valentindavid let me know if you need any help crafting the changes in prepare.sh to have snapd spread tests test this change

mwhudson:
> CC: @mwhudson I do wonder if we should be doing the same on the cdrom device in casper, and if that is the route cause of our troubles. Do we also cleanup udev db between initramfs-tools pivot-root?

I don't see any evidence that we cleanup the db as part of pivoting to the real root. It's possible I'm missing something, but I don't think so... Also I don't think we can be hitting the same sort of issue as this, there aren't any related udev rules trying to ignore synthetic coldplug events are there? (at least i hope not...)

valentindavid:
Note that this change has uncovered some issue with unit dependencies in the initramfs causing some systemd services to start at the wrong moment. So today, I have not followed up with this PR. But I will try to apply the fix in `prepare.sh` tomorrow.

valentindavid:
> nice, this makes a lot of sense, thank you for getting to the bottom of this!
> 
> Before we merge this PR, could you look into applying this patch to the initrd in snapd here: https://github.com/snapcore/snapd/blob/master/tests/lib/prepare.sh#L569-L593
> 
> (essentially just add a line or two which deletes the override file and adds this udev rule to the unpacked initrd)
> 
> and then open a PR over there so we can see all the tests in snapd run with this patch? if things are all happy there I think this patch would then be as well tested as it possibly could be, I'm just nervous that we have missing functionality tests in this repo that are tested inside snapd spread tests and want to see those tests working first.

I have run the spread tests of snapd locally. I have found that one test requires to also have #60 and #61 (with a little fix). The other tests should pass. I have opened draft MR https://github.com/snapcore/snapd/pull/11202 to run the tests in the CI.

valentindavid:
The snapd spread tests have passed along with #59 #60 and #61. See snapcore/snapd#11202

factory/usr/lib/systemd/system/initrd-switch-root.service.d/core-override.conf
```diff
@@ -1,18 +0,0 @@
-[Service]
```
anonymouse64, line 1: can we implement @xnox's suggestion and not delete this file in this PR, just add the rules file, and then in the next PR where you re-work the unit dependencies (https://github.com/snapcore/core-initrd/pull/59) delete this file? AFAICT adding the udev file without deleting this file won't break anything

valentindavid: I have changed this PR. Will put the remaining change in #59

anonymouse64 (approved):
lgtm, thanks for simplifying this PR

mardy (approved):
LGTM!

PR #59, opened 2021-12-01
Rework unit dependencies and use initrd.target as default

valentindavid:
This merge request depends on #58. I will remove it from draft one it will #58 is merged.

This contains 2 commits because I am not sure they can work separately.

##  Rework unit dependencies
    
We should avoid "requiring" or "wanting" units not part of
`initrd-switch-root.target` dependencies. Otherwise they stay when
`initrd-cleanup.service` "isolates" to that target.
    
Having dependencies like `local-fs.target` in
`initrd-switch-root.target` makes "isolate" not stop it. That means
that upon start of the main boot, `local-fs.target` is already active,
making some services to start even though entries in `/etc/fstab` have
not yet been mounted.
    
All our units that stay alive through the switch root should be
"Before" and "WantedBy", directly or indirectly, to any of these
targets: `initrd-root-device.target`, `initrd-root-fs.target`,
`initrd-fs.target`, `initrd.target`.
    
Those units should also have `DefaultDependencies=no` so that they do
not add a "want" or "require" dependency to `sysinit.target`,
`basic.target` or anything those targets depend on. However, they
should probably have `After=sysinit.target` or `After=basic.target`,
directly or indirectly.
    
We avoid dependencies to `initrd-cleanup.service`,
`initrd-switch-root.target` or `initrd-switch-root.service` unless it
is a for a clean up service that needs to run last, like for example
`systemd-bootchart-quit.service`.

## Use initrd target from the beginning
    
Also remove a copy of units to `/etc/systemd/system` as this directory
does not exist and the copy did not seem to be useful.


factory/usr/lib/systemd/system/run-mnt-base.mount.d/fix-dependencies.conf
```diff
@@ -0,0 +1,4 @@
+# TODO, this should probably be generated by `snap-bootstrap
+# initramfs-mounts`.
```
alfonsosanchezbeato, line 2: Maybe it would make sense to create an LP bug affecting snapd and ubuntu-core to keep track of this? So we remove it when it gets implemented there.

alfonsosanchezbeato (commented):
This is awesome, thanks for this! I just have a couple of questions/comments.

factory/usr/lib/systemd/system/populate-writable.service
```diff
@@ -23,4 +27,3 @@ ExecStart=/usr/bin/touch /sysroot/run/fstab
 ExecStart=/usr/bin/chroot /sysroot /usr/lib/core/handle-writable-paths / /etc/system-image/writable-paths /run/fstab
 # TODO:UC20: move the-modeenv implementation to snap-bootstrap too
 ExecStart=/usr/lib/the-modeenv
-ExecStartPost=/usr/bin/systemctl --no-block isolate initrd.target
```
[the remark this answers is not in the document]
valentindavid: What do you mean? The file had no new line, and now it has.

factory/usr/lib/systemd/system/run-mnt-base.mount.d/fix-dependencies.conf
valentindavid: Yes, let's keep this conversation opened until we have a bug opened on launchpad.

factory/usr/lib/the-modeenv
```diff
@@ -20,4 +20,3 @@ if grep -q snapd_recovery_mode=run /proc/cmdline; then
 fi
 mount -o bind /run/image.fstab /sysroot/etc/fstab
 umount -lf /sysroot/run
-cp /run/systemd/transient/*.mount /etc/systemd/system
```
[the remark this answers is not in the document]
valentindavid: Because it copies to `/etc` which is only the ramfs, I believe this has never done anything. If it was copied to a bind mount within `/sysroot/etc`, maybe...

factory/usr/lib/systemd/system/populate-writable.service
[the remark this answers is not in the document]
alfonsosanchezbeato: Oops, github's diff confused me here...

factory/usr/lib/the-modeenv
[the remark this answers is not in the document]
xnox: During `initrd.target` `systemctl daemon-reload` is executed, and at that point, inside the initrd the units from `/run/systemd/transient/` were getting lost as unreferenced.

I think today, snap-bootstrap always generates/copies units in both `/run` and `/etc` such that they survive `daemon-reload` prior to `pivot-root`, thus it is possible that this `cp` call is now redundant.

overall if things still boot correctly, and the mount points are all there, and not recreated, things are good.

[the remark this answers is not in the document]
valentindavid: Are you sure they survived the switch root? Files in `/etc` should not survive. It is files in `/sysroot/etc` that will survive.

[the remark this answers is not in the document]
xnox: This did not survive switch root, but they did survive daemon-reload inside the initrd.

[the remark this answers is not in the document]
valentindavid: Ah, that makes more sense. Though, I am surprised `daemon-reload` loses track of them. It sounds like a bug. I do not see where systemd filters transient units when serializing. I will do some testing.

We still have one `daemon-reload` done by `initrd-parse-etc.service` from systemd. I will see if that loses state.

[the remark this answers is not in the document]
xnox: systemd doesn't really filter transient units, it just `rm -rf` /run/systemd and then reruns generators, which are expected to recreate all the things. Ideally, i was hoping that somehow `systemd-mount` had an option to create mount units persistent i.e. in /etc, instead of in `/run/` but it doesn't really support this.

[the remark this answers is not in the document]
valentindavid: I have tested it, `systemctl daemon-reload` does not remove files in `/run/systemd/transient`, and all the mounts are still listed in `systemctl` (active and mounted). Could it be an older version of systemd that did that?

factory/usr/lib/systemd/system/run-mnt-snapd.mount.d/fix-dependencies.conf
```diff
@@ -0,0 +1,4 @@
+# TODO, this should probably be generated by `snap-bootstrap
+# initramfs-mounts`.
+[Unit]
+DefaultDependencies=no
```
anonymouse64, line 4: can we just have snap-bootstrap do this first without the rest of this PR? Seems like it would be a trivial patch to snap-bootstrap here:

https://github.com/snapcore/snapd/blob/228aa864f305bf6c89d1543b96dc00befe260861/cmd/snap-bootstrap/initramfs_systemd_mount.go#L43-L46

If at all possible, making changes in snap-bootstrap is simpler and quicker and if it doesn't break things in core-initrd can be done first and make changes here simpler/easier to reason about.

anonymouse64: oh actually that unit file is not exactly what we want here that unit file is generated for the other targets to depend on the thing that snap-bootstrap is mounting. It sounds like this would actually be an option to `doSystemdMountImpl` which sets `--property=DefaultDependencies=no` when calling systemd-mount, right?

anonymouse64 (commented):
In general, this PR looks great, it is reducing a lot of the quirkiness of our existing initrd setup and I'm really happy to see that this patch when included in #11202 is successful with passing spread tests, but I do wonder if we couldn't re-order and split out some of this work such that as much as possible is done before we "rip off the bandaid" as it were. I think that the udev rule file from #58 at least could be done without changing to do a stateful re-exec, and it also feels like some of the generation of unit file options for the mounts that snap-bootstrap is doing via systemd-mount could be done without repercussion without the rest of these changes, are there other such changes you can identify that we could split out and do separately?

factory/usr/lib/systemd/system/run-mnt-snapd.mount.d/fix-dependencies.conf
valentindavid: `--property=` is for mount properties. This is a unit property.

valentindavid: Sorry, I have tested, that works.

valentindavid: After some testing, I have figured out, this is not required. Default dependencies added are only After and Before. No new Requires or Wants. So that does not break dependencies for the switch root. I will still do some cleanup in snapd, but that should not change behavior.

valentindavid: I have created snapcore/snapd#11263 but it is strictly required for this PR.

valentindavid:
> In general, this PR looks great, it is reducing a lot of the quirkiness of our existing initrd setup and I'm really happy to see that this patch when included in #11202 is successful with passing spread tests, but I do wonder if we couldn't re-order and split out some of this work such that as much as possible is done before we "rip off the bandaid" as it were. I think that the udev rule file from #58 at least could be done without changing to do a stateful re-exec, and it also feels like some of the generation of unit file options for the mounts that snap-bootstrap is doing via systemd-mount could be done without repercussion without the rest of these changes, are there other such changes you can identify that we could split out and do separately?

@anonymouse64 

I have started to make separate patches. However, I am encountering some difficulties.

So we have a directed graph, and a set of operations (adding or removing an edge) to modify this graph. This set of operations has 2 properties: it does not introduce cycles, and it does not remove certain paths that are important. If we want to split the graph. We can split the set of operations in smaller subsets of operations that from the first graph, creates a new graph where the 2 properties are not broken. However, if I apply 2 subsets I might get to a graph that has broken a property. For example 2 removals can break a path even if each removal does not break it and there is an addition in another subset. 2 additions can introduce a cycle, even if each of the additions are not introducing one and a removal is in a third subset.

In practice I have already encountered such an issue where I had to remove a dependency, but that breaks if I apply another patch after which worked fine by itself.

So I need to make sure that all my patches so that they do not interfere with each others, I will have to test all the orders. And the number of merge orders for the patches is the factorial of the number of patches.

So:
 - I can make multiple PRs that work separately. But once one is merge I will have to retest the next one before we merge.
 - Or I can split the PR in multiple patches to be applied in a given order.

Do you have any preference?

factory/usr/lib/systemd/system/run-mnt-base.mount.d/fix-dependencies.conf
valentindavid: snapcore/snapd#11263

valentindavid:
@anonymouse64 
I have created multiple PRs. Now the PRs to review are  #68 #67 #66 #64 #61 #58. It would be preferable to merge them in that order since I have tested all of them rebased in that order in #69. (By tested I mean, install secure boot, create account, login, check journal for errors).

Once they are all merged, the last commit to finally enable stateful re-execution is in #70.

factory/usr/lib/systemd/system/run-mnt-snapd.mount.d/fix-dependencies.conf
anonymouse64: thanks for looking into that, I reviewed that PR

anonymouse64:
@valentindavid the specific order to land things is fine as long as you do as you did in your comment and tell us the order that things should be landed in. Thanks for splitting this up and documenting the order things should land in

anonymouse64:
oh also if you need to land things in a specific order and they depend on each other, it is okay to file PR's like this:
1. PR 1 has commit `abc`
2. PR 2 depends on `abc` and has commits `abc` and `foo`
3. PR 3 depends on `abc` and `foo` and has commits `abc` and `foo` and `xyz`
4. ...

we do that in snapd a lot, just again the thing is to mention the ordering of things so it's clear to reviewers, also while not necessary it is helpful in say PR 3 to mention that it is based on top of PR 2 and that in that PR the main commit to review is `xyz` etc.

valentindavid:
All changes have been put in other PRs. I am closing this one.

PR #60, opened 2021-12-01
Kill recovery-chooser-trigger "gracefully" so it can restart

valentindavid:
This is a follow up for #58 and will be draft until #58 is merged.

`recovery-chooser-trigger` will be stopped when isolating to target
`initrd-switch-root.target`. The problem is that the exit status of
the process will be considered a failure. During main boot,
`recovery-chooser-trigger` will not be restarted because the stateful
reexecution will remember the failure.
    
The long term solution will be the make `recovery-chooser-trigger`
survive the switch root instead.


xnox:
One can use `@` prefix of the recovery-chooser-trigger Exec= line, which should make it continue running from initrd post pivot.

This is what mdmon does. Is that what is desired here? Then something eventually should really stop it, once we booted post pivot.

valentindavid:
> One can use `@` prefix of the recovery-chooser-trigger Exec= line, which should make it continue running from initrd post pivot.

I suppose you mean `ExecStart=@/usr/lib/snapd/snap-bootstrap @ recovery-chooser-trigger`. I can try.

xnox:
> > One can use `@` prefix of the recovery-chooser-trigger Exec= line, which should make it continue running from initrd post pivot.
> 
> I suppose you mean `ExecStart=@/usr/lib/snapd/snap-bootstrap @ recovery-chooser-trigger`. I can try.

`ExecStart=@/usr/lib/snapd/snap-bootstrap recovery-chooser-trigger`

That yes.

valentindavid:
> ExecStart=@/usr/lib/snapd/snap-bootstrap recovery-chooser-trigger

According to systemd.service's man page:
> If the executable path is prefixed with "@", the second specified token will be passed as "argv[0]" to the executed process (instead of the actual filename), followed by the further arguments specified.

Which means argv will be just `{"recovery-chooser-trigger"}`. Which would be failing.

According to http://systemd.io/ROOT_STORAGE_DAEMONS/ we need to have `argv[0][0] == '@'`.

Which means we would need: `ExecStart=@/usr/lib/snapd/snap-bootstrap @ recovery-chooser-trigger`

xnox:
> > ExecStart=@/usr/lib/snapd/snap-bootstrap recovery-chooser-trigger
> 
> According to systemd.service's man page:
> 
> > If the executable path is prefixed with "@", the second specified token will be passed as "argv[0]" to the executed process (instead of the actual filename), followed by the further arguments specified.
> 
> Which means argv will be just `{"recovery-chooser-trigger"}`. Which would be failing.
> 
> According to http://systemd.io/ROOT_STORAGE_DAEMONS/ we need to have `argv[0][0] == '@'`.
> 
> Which means we would need: `ExecStart=@/usr/lib/snapd/snap-bootstrap @ recovery-chooser-trigger`

recovery-chooser-trigger is an argv2 to snap-bootstrap which is the argv0....

valentindavid:
That worked. I have created a new PR on snapd that will test that along with some other improvements. I will open a PR here if that works.

valentindavid:
This was tested by snapd's CI and worked. I have opened it as #64.

PR #61, opened 2021-12-01
Make sure tmp.mount is stopped before switch

valentindavid:
This is a follow-up to #58. It will be unmarked as draft once #58 is merged.

Otherwise with stateful re-execution, systemd will think for a short
time that it is mounted. It will start `systemd-tmpfiles-setup` and
`systemd-timesyncd` too early.

### DIFF
--- a/factory/usr/lib/systemd/system/tmp.mount.d/umount.conf
+++ b/factory/usr/lib/systemd/system/tmp.mount.d/umount.conf
@@ -1,0 +1,6 @@
+# Mounts do not seem to be stopped when isolating, so we need to force
+# a conflict to umount tmp.mount so that main boot does not think it
+# is already mounted from the state.
+[Unit]
+Conflicts=initrd-switch-root.target
+Before=initrd-switch-root.target


anonymouse64:
Just a comment that we need to be extra robust in our testing of things that might change how systemd-timesyncd works, because we have had numerous bugs before around devices without a RTC getting the wrong time in the initrd and the boot breaks because snapd can't validate assertions that are in the future and snapd (due to the wrong time) thinks the assertions are in the future.

valentindavid:
> Just a comment that we need to be extra robust in our testing of things that might change how systemd-timesyncd works, because we have had numerous bugs before around devices without a RTC getting the wrong time in the initrd and the boot breaks because snapd can't validate assertions that are in the future and snapd (due to the wrong time) thinks the assertions are in the future.

systemd-timesyncd is started too early with #52. This commit changes the behavior to start like before #52 is applied.

xnox:
As per expectations only /run is persisted between initrd and running system. Anything that relied on files in /tmp to be available across initrd & running system was programmed incorrectly and might need fixing. But also was never expected to work anyway.

valentindavid:
Nothing expects /tmp to be persisted.

valentindavid:
An alternative is to remove `tmp.mount` from initramfs alltogether. It does not seem to be used by any service in initramfs.

valentindavid:
> An alternative is to remove `tmp.mount` from initramfs alltogether. It does not seem to be used by any service in initramfs.

I have tested that, and it seems to be working without `tmp.mount`.

Note that dracut for example does not add `tmp.mount`.

alfonsosanchezbeato:
@valentindavid so, do you think we should fully remove `tmp.mount`? Or you think that would be too risky and we should keep this PR as it is now?

alfonsosanchezbeato (approved):
LGTM, thanks

Meulengracht (approved):
LGTM!

PR #63, opened 2022-01-03
Add support for lxd-backend

Meulengracht:
This is the same as I've done for core20 and core-base. This PR adds support for the LXD backend, adds instructions on how to run both the qemu and lxd backend, and it switches our CI to self-hosted so the CI should work again.

This also pulls in our shared testing repostiory as a subtree, which unfortunately muddies the PR a bit and I'm sorry for the size.


Meulengracht:
Closing this for now as self-hosted runners for core-initrd has yet to be setup

PR #67, opened 2022-01-17
Rework dependencies of mount units

valentindavid:
Mount units that survive the switch root should not use default
dependencies. Instead they should be part of either `initrd-fs.target`
or `initrd-root-fs.target`.

This removes "Wants"/"Requires" dependencies from the surviving mounts
to units that do not need to survive the switch root so they do not
pull dependencies into the swith root.

So mount units should not "require" nor "want" `the-tool.service`.
They only need "After". The "Wants" to `the-tool.service` is done
`basic.target` (and should be moved to `initrd.target` when the
default target hack is removed). The same is needed for
`populate-writable.service` for which this change introduce a "Wants"
dependency from `basic.target`.


anonymouse64:
before I keep reviewing these, since I see you split these up so they are independent and don't depend on each other, just let me know if it would be simpler/easier for you to rebase PR's on top of each other as I explained in https://github.com/snapcore/core-initrd/pull/59#issuecomment-1016850897

valentindavid:
> before I keep reviewing these, since I see you split these up so they are independent and don't depend on each other, just let me know if it would be simpler/easier for you to rebase PR's on top of each other as I explained in [#59 (comment)](https://github.com/snapcore/core-initrd/pull/59#issuecomment-1016850897)

I have split most of PRs so they can be reviewed separately. If you want to see how they will look once rebased in order, look at #69. And #70 contains all the changes.

alfonsosanchezbeato (approved):
LGTM

anonymouse64 (approved):
lgtm

PR #68, opened 2022-01-18
Use initrd target from the beginning

valentindavid:
Also remove a copy of units to `/etc/systemd/system` as this directory
does not exist and the copy did not seem to be useful.


factory/usr/lib/systemd/system/initrd-root-fs.target.d/42-kernel.conf
```diff
@@ -1,5 +1,4 @@
 [Unit]
-Requires=run-mnt-kernel.mount
```
anonymouse64, line 2: just to be clear, the Requires here are being removed because they are superfluous as you mention in your other comment?

factory/usr/lib/the-tool
```diff
@@ -4,14 +4,3 @@ set -e
 # Run snap-bootstrap
 /usr/lib/snapd/snap-bootstrap initramfs-mounts
 
-# TODO:UC20: move this code into snap-bootstrap proper so this script _just_
```
anonymouse64, line 7: this doesn't have to be done as part of this PR and in fact can be done after your chain of PR's have all been landed, but it would be nice to get rid of this wrapper script as mentioned

anonymouse64 (approved):
lgtm, do we know why the tests are failing/not running on this PR?

factory/usr/lib/systemd/system/initrd-root-fs.target.d/42-kernel.conf
valentindavid: Those do not exist at boot. They are generated by `the-tool.service`. In practice I have noticed it tends to confuse systemd because the graph of jobs is built at load. So generating new units requires `daemon-reload`. Instead we can just make sure to have `Wants=the-tool.service` and `After=the-tool.service` and remove the mentions to `run-mnt-*.mount`. We only need to keep them in the `Wants` of the initrd targets.

factory/usr/lib/the-tool
valentindavid: Yes, I have noted that. I will come back to it. This is just not needed for stateful re-execution so I will do it after.

Meulengracht (approved):
Just a question, however otherwise the order looks fine to me, and it I had no problems with testing - so hopefully we can get CI verifying quickly as well!

factory/usr/lib/systemd/system/initrd-root-device.target.d/42-kernel.conf
```diff
@@ -1,4 +1,3 @@
 [Unit]
-Requires=run-mnt-kernel.mount
```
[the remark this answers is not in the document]
valentindavid: Here we remove the Requires to make the boot more robust. There seems to be issues in ordering when requiring units that do not exist yet. This unit is created later by `the-tool.service`.

There is another PR that cleans up more the graph and removes this file. That because really `initrd-root-device.target` has no reason to depend on `run-mnt-kernel.mount`.

But I wanted to keep this PR minimal, I just had to remove the Requires to make systemd order start jobs properly.

alfonsosanchezbeato (approved):
LGTM!

Meulengracht (approved):

factory/usr/lib/systemd/system/initrd-root-fs.target.d/42-kernel.conf
anonymouse64: Thanks for confirming

anonymouse64 (approved):
still lgtm

PR #72, opened 2022-01-25
spread: add spread and CI testing for core20 version of core-initrd

Meulengracht:
This is an identical PR to the one done in core20 PR that adds the 'spread lxd-nested' testing backend, and updates documentation on how to test with qemu backend as well. 

This also adds the snapd-testing-tools repository (so sorry for the size), and switches our CI to self-hosted runners that should enable PRs to run tests. I have also enabled testing for the core20 branch.

I will create a PR for the main branch as well that is modified to work with jammy.


tests/lib/prepare-uc20.sh
```diff
@@ -258,26 +258,46 @@ echo "ubuntu:x:1001:" >> /root/test-var/lib/extrausers/group
...
-    /root/test-etc /root/test-var/lib/extrausers
-
-# tear down the mounts
-umount /mnt
-kpartx -d pc.img
+# for the lxd backend this step is a bit different as the device mapper kernel module
+# is not supported by lxd containers. We thus have to do some manual setup of the image
+# partition mount.
+if [ "${SPREAD_BACKEND}" = "lxd-nested" ]; then
+    devloop=$(losetup -f)
```
valentindavid, line 262: You can use `-f` directly into the `losetup` command 2 lines under. That will save us from a potential race where something takes that loop device.

tests/lib/external/snapd-testing-tools/.github/workflows/tests.yaml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Run shellCheck for tools
+      run: |
+          sudo apt-get remove --purge shellcheck
+          sudo snap install shellcheck
+
+          find tools -type f -exec sh -c "head -n 1 {} | egrep -a 'bin/bash|bin/sh' >/dev/null" \; -print -exec shellcheck {} \;
```
valentindavid, line 22: To not fail on file names with spaces, that should work  better:

`find tools -type f -exec sh -c 'head -n 1 "${1}" | egrep -a "bin/bash|bin/sh" >/dev/null' '--' {} \; -print -exec shellcheck {} \;`

tests/lib/external/snapd-testing-tools/COPYING
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
valentindavid, line 1: I see this commit imports some shared code from somewhere else. What is the procedure to keep track of updates?

Meulengracht: As we talked about the code in that folder is a subtree

tests/lib/prepare-uc20.sh
Meulengracht: Ah nice, i did not know about this, i'll update it!

valentindavid (commented):
I see that 527b358ec62fd5e06c791ee668c9fab75325d347 adds files later removed by 54d9d03ffeff18ad76a9da95c2d8099887a53ad4. Could you remove those commits from the history. Also 82d1fad0937589b32816379776cd13717c54673d is empty, please remove it.

valentindavid (commented):
Why did 132e13a2a76991f615ed9fdae0a5b0765c83c22f remove `tests/lib/nested.sh` and b8b0c923af2c2c50376fb49f52db2bb4db946b4b re-added modified? It makes it difficult to review.

tests/lib/nested.sh
```diff
@@ -0,0 +1,210 @@
...
+    printf '[Unit]\nDescription=QEMU Nested VM for UC20 spread job %s\n[Service]\nType=simple\nExecStart=%s\n' \
+        "${SPREAD_JOB:-unknown}" \
+        "${SVC_SCRIPT}" \
+            > "/run/systemd/system/${SVC_NAME}.service"
+    
+    systemctl daemon-reload
+    systemctl start "${SVC_NAME}"
+
+    # wait for the nested-vm service to appear active
+    if ! wait_for_service "${SVC_NAME}"; then
```
valentindavid, line 202: Any reason why not just using `systemctl start --wait "${SVC_NAME}"`?

tests/lib/nested.sh
```diff
@@ -0,0 +1,210 @@
...
+    PARAM_SERIAL="-serial file:${WORK_DIR}/serial.log"
+    PARAM_TPM=""
+
+    ATTR_KVM=""
+    if [ "$ENABLE_KVM" = "true" ]; then
+        ATTR_KVM=",accel=kvm"
+        # CPU can be defined just when kvm is enabled
+        PARAM_CPU="-cpu host"
+        # Increase the number of cpus used once the issue related to kvm and ovmf is fixed
+        # https://bugs.launchpad.net/ubuntu/+source/kvm/+bug/1872803
```
valentindavid, line 125: On 2020-06-22, there was a comment asking to try again because a fix was done. Has anyone tried to follow up?

tests/lib/nested.sh
```diff
@@ -0,0 +1,210 @@
...
+    fi
+
+    mkdir -p "${WORK_DIR}/image/"
+    cp -f "/usr/share/OVMF/OVMF_VARS${OVMF_VARS}.fd" "${WORK_DIR}/image/OVMF_VARS${OVMF_VARS}.fd"
+    PARAM_BIOS="-drive file=/usr/share/OVMF/OVMF_CODE${OVMF_CODE}.fd,if=pflash,format=raw,unit=0,readonly -drive file=${WORK_DIR}/image/OVMF_VARS${OVMF_VARS}.fd,if=pflash,format=raw"
+    PARAM_MACHINE="-machine q35${ATTR_KVM} -global ICH9-LPC.disable_s3=1"
+
+    if [ "${ENABLE_TPM:-false}" = "true" ]; then
+        TPMSOCK_PATH="/var/snap/swtpm-mvo/current/swtpm-sock"
+        if [ "${SPREAD_BACKEND}" = "lxd-nested" ]; then
```
valentindavid, line 146: It is not obvious to me why we use a snap in some cases and not others. This is the kind of thing that needs a comment.

tests/lib/nested.sh
```diff
@@ -0,0 +1,210 @@
...
+        PARAM_TPM="-chardev socket,id=chrtpm,path=${TPMSOCK_PATH} -tpmdev emulator,id=tpm0,chardev=chrtpm -device tpm-tis,tpmdev=tpm0"
+    fi
+
+    PARAM_IMAGE="-drive file=${IMAGE_FILE},cache=none,format=raw,id=disk1,if=none -device virtio-blk-pci,drive=disk1,bootindex=1"
+
+    # Create the systemd unit for running the VM, the qemu parameter order is 
+    # important. We run all of the qemu logic through a script that systemd runs
+    # for an easier time escaping everything
+    SVC_NAME="nested-vm-$(systemd-escape "${SPREAD_JOB:-unknown}")"
+    SVC_SCRIPT="/tmp/nested-vm-$RANDOM.sh"
```
valentindavid, line 171: Why not using `mktemp`?

tests/lib/nested.sh
```diff
@@ -0,0 +1,210 @@
...
+    ${PARAM_SERIAL} \
+    ${PARAM_MONITOR}
+EOF
+    
+    chmod +x "${SVC_SCRIPT}"
+
+    printf '[Unit]\nDescription=QEMU Nested VM for UC20 spread job %s\n[Service]\nType=simple\nExecStart=%s\n' \
+        "${SPREAD_JOB:-unknown}" \
+        "${SVC_SCRIPT}" \
+            > "/run/systemd/system/${SVC_NAME}.service"
```
valentindavid, line 196: Why not `systemd-run`?

tests/lib/nested.sh
```diff
@@ -0,0 +1,210 @@
...
+
+execute_remote(){
+    sshpass -p ubuntu ssh -p "$SSH_PORT" -o ConnectTimeout=10 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ubuntu@localhost "$*"
+}
+
+wait_for_service() {
+    local service_name="$1"
+    local state="${2:-active}"
+    for i in $(seq 300); do
+        if systemctl show -p ActiveState "$service_name" | grep -q "ActiveState=$state"; then
```
valentindavid, line 16: `systemctl is-active`?

tests/lib/nested.sh
```diff
@@ -0,0 +1,210 @@
...
+    echo "service $service_name did not start"
+    exit 1
+}
+
+wait_for_ssh(){
+    local service_name="$1"
+    retry=800
+    wait=1
+    while ! execute_remote true; do
+        if ! systemctl show -p ActiveState "$service_name" | grep -q "ActiveState=active"; then
```
valentindavid, line 35: `systemctl is-active`?

HACKING.md
```diff
@@ -23,6 +23,86 @@ TODO: Explain how to modify initrd when there is Secure boot / FDE.
...
+ - Spread (install from source as per below)
+
+## Installing spread
+
+You can install spread by simply using ```snap install spread```, however this does not allow for the lxd-backend to be used.
+To use the lxd backend you need to install spread from source, as the LXD profile support has not been upstreamed yet.
+This document will be updated with the upstream version when this happens. To install spread from source you need to do the following.
+
+```
+git clone https://github.com/Meulengracht/spread
```
valentindavid, line 41: Could we maybe comment that we need to change the document when `https://github.com/snapcore/spread/pull/136` is merged?

HACKING.md
```diff
@@ -23,6 +23,86 @@ TODO: Explain how to modify initrd when there is Secure boot / FDE.
...
+the qemu backend. This is because the container can use all the resources of the host, and we can support
+qemu-kvm acceleration in the container for the nested instance.
+
+This backend requires that your host machine supports KVM.
+
+1. Setup any prerequisites and build the LXD image needed for testing. The following commands will install lxd
+and yq (needed for yaml manipulation), download the newest image and import it into LXD.
+```
+sudo snap install lxd --channel=latest/stable
+sudo snap install yq --channel=latest/stable
```
valentindavid, line 81: Is `--channel=latest/stable` really needed?

HACKING.md
```diff
@@ -23,6 +23,86 @@ TODO: Explain how to modify initrd when there is Secure boot / FDE.
...
+curl -o lxd-initrd-img.tar.gz https://storage.googleapis.com/snapd-spread-core/lxd/lxd-spread-initrd-img.tar.gz
```
valentindavid, line 82: Just a question, what builds and publishes this image? I would like to know how to generate a new one.

Also why not using `-O` instead of `-o lxd-initrd-img.tar.gz` to make the command line shorter? It is nice to have concise lines in a document.

valentindavid: Other thing, that URL did not work for me.

tests/lib/nested.sh
Meulengracht: I believe there is a lot of code left-overs as most of this is adapted from the snapd repository, which could explain the wait_for_service being flexible, even though we do not use any other state than active in this script. I'll try to run it with your suggestions and see if it runs correctly

tests/lib/nested.sh
Meulengracht: It would be nice to still have it named something easy to recognize so if you have to debug the script it generates, or wanting to inspect how the inner VM is started. Does this make sense or would you still prefer it being mktemp?

tests/lib/nested.sh
Meulengracht: I assume it's to support multiple different states, and not just active - however it's not something that is required for this instance, so I think we might just go with your suggestions here

HACKING.md
Meulengracht: Yes, definiately a good suggestion

HACKING.md
Meulengracht: Good catch

tests/lib/nested.sh
valentindavid: You can give a template. `mktemp --tmpdir nested-vm-XXXXXXXXXX`

HACKING.md
Meulengracht: Ok so this is a bit hen-and-egg issue, I needed it to run from the snapcore repo, and the image is built by the github workflow. So I kind of need it merged, and then i'll start the workflow so the image is generated. I've been generating it locally while testing.

tests/lib/nested.sh
Meulengracht: Cool, i'll test it out!

tests/lib/nested.sh
Meulengracht: I'll try this out and see if it crashes!

HACKING.md
```diff
@@ -23,6 +23,86 @@ TODO: Explain how to modify initrd when there is Secure boot / FDE.
...
+lxc image show ucspread > temp.profile
+yq e '.properties.aliases = "ucspread,amd64"' -i ./temp.profile
+yq e '.properties.remote = "images"' -i ./temp.profile
+cat ./temp.profile | lxc image edit ucspread
+rm ./temp.profile ./lxd-initrd-img.tar.gz
+```
+2. Import the LXD coreinitrd test profile. Make sure your working directory is the root of this repository.
+```
+lxc profile create coreinitrd
+cat tests/spread/core-initrd.lxdprofile | lxc profile edit coreinitrd
```
valentindavid, line 93: I did not find file `tests/spread/core-initrd.lxdprofile`

Meulengracht: Yea I added it yesterday in a later commit, I had lost it during the commit cleanup, can you try to pull the newest?

tests/lib/prepare-uc20.sh
```diff
@@ -257,26 +255,49 @@ echo "ubuntu:x:1001:" >> /root/test-var/lib/extrausers/group
...
+
+    # add the data that snapd.spread-tests-run-mode-tweaks.service reads to the 
+    # mounted partition
+    tar -c -z \
+        -f /mnt/run-mode-overlay-data.tar.gz \
+        /root/test-etc /root/test-var/lib/extrausers
+
+    # tear down the mounts
+    umount /mnt
+    kpartx -d pc.img
```
alfonsosanchezbeato, line 300: It would be nice to do this also in a term trap to make sure the unmounting happens also if something fails between mounting and unmounting:

    f() {
        umount...
    }
    trap f TERM

But this is just a suggestion, if this happens in a temporary container then it does not matter too much.

valentindavid: or `trap f EXIT` so that it works not only for `SIGTERM`.

Meulengracht: It is executed either in the LXD container or a VM - so it should not cause any issues, however I still believe we should apply this as it would be more proper. Thanks for your suggestions alfonso, greatly appreciated

PR #73, opened 2022-01-26
spread: add spread and CI testing

Meulengracht:
This PR adds spread testing to the jammy version of core-initrd. It adds CI testing, LXD/Qemu backend and documentation on how to set it up. 

It should be more or less identical in structure to the earlier spread PRs, however  this one has quite a few changes. The image is a lot different, and the number of packages and how we install things have changed. There is an issue with installing snaps in privileged containers which results in a failing of reloading udev rules (see https://bugs.launchpad.net/snapd/+bug/1712808).

The workaround in this testing is to try the installation again, which then succeeds.


valentindavid (approved):

alfonsosanchezbeato (approved):
LGTM!

PR #74, opened 2022-02-09
Remove writable-paths

valentindavid:
Instead we expect to load mounts from /etc/fstab from core.

/sysroot /sysroot/writable are the only mountpoints that need to be
mounted before /etc/fstab loaded, so only those mount units are left.
The rest will be automatically loaded.

Before loading /etc/fstab we need to prepare the writable disk. So
before initrd-root-fs.target and after /sysroot/writable is mounted,
we call systemd-tmpfiles for core-writable.conf.

This commit requires change in core-base to provide
/etc/fstab and core-writable.conf.

See snapcore/core-base#25


factory/usr/lib/systemd/system/initrd-core-system-defaults.service
```diff
@@ -0,0 +1,16 @@
...
+ConditionKernelCommandLine=|snapd_recovery_mode=install
+ConditionKernelCommandLine=|snapd_recovery_mode=recover
+After=initrd-root-fs.service
+Before=initrd-fs.service
+RequiresMountsFor=/run/mnt/data
+
+[Service]
+Type=oneshot
+ExecStart=-/bin/mkdir -p /run/mnt/data/system-data/etc
+ExecStart=/bin/cp -aT /run/mnt/data/system-data/_writable_defaults/etc /sysroot/writable/system-data/etc
```
alfonsosanchezbeato, line 14: How was this done before this PR?

valentindavid: Somewhere hidden in `handle-writable-paths` with a TODO to move it into snap-bootstrap.

factory/usr/lib/systemd/system/sysroot-writable.mount
```diff
@@ -1,12 +1,13 @@
...
-
-Wants=sysroot.mount
-After=sysroot.mount
-
-After=run-mnt-data.mount
-After=run-mnt-base.mount
+Before=initrd-root-fs.target
+RequiresMountsFor=/writable
+RequiresMountsFor=/run/mnt/data
+Wants=the-tool.service
```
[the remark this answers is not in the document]
valentindavid: I will need to rebase.

factory/usr/lib/systemd/system/initrd-core-system-defaults.service
valentindavid: Ah no sorry. I misread.

valentindavid: Actually that was also a part of `handle-writable-paths` in function `handle_writable_defaults`.

factory/usr/lib/systemd/system/sysroot-writable.mount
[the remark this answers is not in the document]
valentindavid: I have rebased.

factory/usr/lib/systemd/system/initrd-core-system-defaults.service
alfonsosanchezbeato: I see it now, thanks for the pointer

pedronis:
As we hinted/discussed and after talking with @mvo5 again, given the timing, at this point this change should be scheduled for UC 24, to be done early in its development

anonymouse64 (commented):
the approach here has my +1 but I unfortunately won't have time to review the details

xnox:
And here we are in UC24. Can we do it? is it already done elsehow?

pedronis:
@xnox
>  And here we are in UC24. Can we do it? is it already done elsehow?

we didn't do it early in the development, so not yet

valentindavid:
Moved to https://github.com/canonical/snapd/pull/15775

PR #75, opened 2022-02-09
Remove tmp.mount

valentindavid:
/tmp is not useful in initrd for any of its service. It also requires
some hack to unmount properly. It is better to get rid of it.


anonymouse64 (approved):
lgtm thanks for this

xnox (approved):

PR #77, opened 2022-02-10
tests: Be more verbose when nested VM does not respond

valentindavid:
This prints the output of the serial port if we were unable to
connect through ssh to the VM.


alfonsosanchezbeato (approved):
LGTM, thanks

PR #76, opened 2022-02-09
Rename "the-tool" and remove wrapper script

valentindavid:

PR #78, opened 2022-02-10
tests: Properly call ubuntu-core-initramfs to generate

valentindavid:
The script was merging manually the skeleton files into a tree of an
old build of core-initrd. It would fail when files needed to be
removed.

Instead we need to make a fresh initramfs and embed it into the
kernel.efi.


tests/lib/prepare-uc22.sh
```diff
@@ -188,19 +188,23 @@ rm -r $snapddir
...
-# repack the initrd into the kernel.efi
-"$EXTTESTSLIB/repack-kernel" prepare $kerneldir
+unsquashfs -f -d "${kerneldir}" upstream-pc-kernel.snap
+(
+    cd "${kerneldir}"
+    config="$(echo config-*)"
+    kernelver="${config#config-}"
+    objcopy -O binary -j .linux kernel.efi kernel.img-"${kernelver}"
+    ubuntu-core-initramfs create-initrd --kerneldir modules/"${kernelver}" --kernelver "${kernelver}" --firmwaredir firmware --output ubuntu-core-initramfs.img
+    ubuntu-core-initramfs create-efi --initrd ubuntu-core-initramfs.img --kernel kernel.img --output kernel.efi --kernelver "${kernelver}"
```
xnox, line 201: Should it be `--kernel kernel.img-"${kernelver}"` ? given that's where the kernel image got extracted to?

valentindavid: `        args.kernel = "-".join([args.kernel, args.kernelver])`

valentindavid:
This change is needed for #76

tests/lib/prepare-uc22.sh
xnox: How confusing! Especially since it's probably me who wrote that.

Meulengracht (approved):

alfonsosanchezbeato:
LGTM, thanks! Probably this one should be backported to the `core20` branch.

alfonsosanchezbeato (approved):

PR #80, opened 2022-02-14
[core20] tests: Properly call ubuntu-core-initramfs to generate the initrd

valentindavid:
This is a backport of #78

The script was merging manually the skeleton files into a tree of an
old build of core-initrd. It would fail when files needed to be
removed.

Instead we need to make a fresh initramfs and embed it into the
kernel.efi.


alfonsosanchezbeato (approved):
LGTM, thanks for backporting

PR #81, opened 2022-02-14
factory: sync users/groups with core22

alfonsosanchezbeato:
Sync users and groups with current ones for the core22 snap. This
removes this error from the journal:

systemd-udevd[<pid>]: /usr/lib/udev/rules.d/50-udev-default.rules:42 Unknown group 'sgx', ignoring


xnox (commented):
Removal of docker group sounds odd. Given that i thought we must have it hardcoded to keep docker snap working on Ubuntu Core / Classic. Is it a oversight that docker group went missing in core22 versus core18 or core20?

alfonsosanchezbeato:
I'm not sure if that was an oversight, there was this PR in core-base that made sure to avoid touching GID 113: https://github.com/snapcore/core-base/pull/11
There was also a try to create docker uid/gid in extrausers, that was reverted: https://github.com/snapcore/core20/pull/87
But I do not see docker user/group in core20 snap either. I do not know what is creating that user/group, in case it is being created at all. Maybe @anonymouse64 or @sil2100 can help with clarifying this.

xnox:
> I'm not sure if that was an oversight, there was this PR in core-base that made sure to avoid touching GID 113: [snapcore/core-base#11](https://github.com/snapcore/core-base/pull/11) There was also a try to create docker uid/gid in extrausers, that was reverted: [snapcore/core20#87](https://github.com/snapcore/core20/pull/87) But I do not see docker user/group in core20 snap either. I do not know what is creating that user/group, in case it is being created at all. Maybe @anonymouse64 or @sil2100 can help with clarifying this.

weird.

docker stuff has been flip-flopping. Some definitive input from like @anonymouse64 @pedronis @mvo5 would be helpful. Maybe docker stuff is not handled via newer user/daemon/group stuff and indeed no longer needed in neither core-initrd nor in base snaps, for neither classic nor core.

xnox (approved):
I am happy for this to be merged; as long as docker group presence/removal/requirements are double-checked in a follow up.

anonymouse64:
This is correct - there should not be a docker group defined in the base snaps of Ubuntu Core 20 and Ubuntu Core 22. That's all we need right now, we don't have system-usernames or any other method, but that's "okay"

PR #86, opened 2022-02-25
Backport patches from sd-stub v250

alfonsosanchezbeato:
Backport patches for systemd stub from systemd v250. These patches add
support to arm64 kernels by using a new kernel feature [1] for passing
the address of the initramfs to the kernel stub by a special device
path using the LoadFile2 EFI protocol. This feature is now used also
for x86. Support is in kernel 5.7 for arm64 and in 5.8 for x86.

The patches included are those affecting the src/boot/efi/ folder up
to commit [2]. Note that we are patching the Ubuntu package, which
uses as base
https://github.com/systemd/systemd-stable/releases/tag/v249.10 , so
some patches from v250 are already backported there. Additionally,
[3] from post-v250 has been included as well.

When backporting the patches, [4] has not been included because we
need to set a PCR different to the one in grub for measuring kernel
command line. An issue for this problem [5] has been opened in systemd
upstream. Also, [6] was not included either because it touched many
paths outside src/boot/efi/ and it is a cosmetic change.

Finally, some commits that do not touch src/boot/efi/ were also
included as they were needed because they do some refactoring of
macros that affect the sd-stub code. These are:

0003-alloc-util-make-mfree-typesafe.patch 5afcf89ca29b518eb2fa244b015afc2708f77e1d
0004-macro-Move-some-macros-to-macro-fundamental.h.patch f862e847246b5588d9d6ed7d91da11b6adbf39e7
0069-fundamental-define-size_t-and-memcpy-for-sd-boot.patch 5d8a725b0800ce572bb3308c03e98561251a9284
0071-move-mfree-to-macro-fundamentals.h.patch 200b1d997d96179aba9489ce9d373e869557460e

[1] https://github.com/torvalds/linux/commit/ec93fc371f014a6fb483e3556061ecad4b40735c
[2] https://github.com/systemd/systemd/commit/33bc9b756ea9a53c31340a7dcc6bc6b45c0fb2f5
[3] https://github.com/systemd/systemd/commit/178d598b5fae36fa9d54c30668771f9c626724f6
[4] https://github.com/systemd/systemd/commit/faacf1807e8fcbee4bf60495e0145e4522d76393
[5] https://github.com/systemd/systemd/issues/22635
[6] https://github.com/systemd/systemd/commit/fce9abb22793505fbb09eccd0b1e24b5d578dbcc

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -7,6 +7,8 @@ include /usr/share/dpkg/default.mk
 	dh $@ --buildsystem=meson+ninja --sourcedirectory=vendor/systemd --with python3
 
 override_dh_auto_configure:
+	find vendor/systemd-patches/ -type f -printf "%f\n" | sort >> vendor/systemd/debian/patches/series
+	cp vendor/systemd-patches/*.patch vendor/systemd/debian/patches/
 	cd vendor/systemd; QUILT_PATCHES=debian/patches quilt push -a || true
 	sed -i '/"minix"/s/^/\/\//' vendor/systemd/src/test/test-path-util.c
 	dh_auto_configure --buildsystem=meson+ninja --sourcedirectory=vendor/systemd -- \
--- a/vendor/systemd-patches/0001-Disable-non-explicit-sbatvars-autodetection-for-cros.patch
+++ b/vendor/systemd-patches/0001-Disable-non-explicit-sbatvars-autodetection-for-cros.patch
@@ -1,0 +1,30 @@
+From 356320ac48c558a8b08bd7f22a80685791a8f04f Mon Sep 17 00:00:00 2001
+From: James Hilliard <james.hilliard1@gmail.com>
+Date: Sat, 10 Jul 2021 04:36:50 -0600
+Subject: [PATCH 01/73] Disable non-explicit sbatvars autodetection for cross
+ builds.
+
+Since autodetection is unlikely to work reliably for cross builds
+disable it unless explicitly enabled.
+
+Signed-off-by: James Hilliard <james.hilliard1@gmail.com>
+---
+ src/boot/efi/meson.build | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/src/boot/efi/meson.build b/src/boot/efi/meson.build
+index afdf739d9..62d826bec 100644
+--- a/src/boot/efi/meson.build
++++ b/src/boot/efi/meson.build
+@@ -115,7 +115,7 @@ if have_gnu_efi
+                             ['sbat-distro-url', 'BUG_REPORT_URL']]
+                 foreach sbatvar : sbatvars
+                         value = get_option(sbatvar[0])
+-                        if value == '' or value == 'auto'
++                        if (value == '' and not meson.is_cross_build()) or value == 'auto'
+                                 cmd = 'if [ -e /etc/os-release ]; then . /etc/os-release; else . /usr/lib/os-release; fi; echo $@0@'.format(sbatvar[1])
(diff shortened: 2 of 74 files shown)


valentindavid:
Is there an issue opened on systemd to track revert/change of https://github.com/systemd/systemd/commit/faacf1807e8fcbee4bf60495e0145e4522d76393 ?

Are all the patches really required? Or was it more convenient to provide all change in `src/boot/efi` because refactoring makes it difficult rebase fixes?

alfonsosanchezbeato:
> Is there an issue opened on systemd to track revert/change of [systemd/systemd@faacf18](https://github.com/systemd/systemd/commit/faacf1807e8fcbee4bf60495e0145e4522d76393) ?

I've opened it now: https://github.com/systemd/systemd/issues/22635

> 
> Are all the patches really required? Or was it more convenient to provide all change in `src/boot/efi` because refactoring makes it difficult rebase fixes?

It was more convenient - `sd-{boot,stub}` has been severaly refactored, so it was easier to just take all the patches instead of backporting what we needed.

alfonsosanchezbeato:
> > > Are all the patches really required? Or was it more convenient to provide all change in `src/boot/efi` because refactoring makes it difficult rebase fixes?
> > 
> > 
> > It was more convenient - `sd-{boot,stub}` has been severaly refactored, so it was easier to just take all the patches instead of backporting what we needed.
> 
> How did you select what commit to backport? Did you select manually? Or did you filter by commits modifying specific files?

I took any commit modifying `src/boot/efi/` up to the point where using EFI for passing the initrd is implemented, although some of the patches also modify bits in other places (some refactoring of where macros live).

alfonsosanchezbeato:
The following ones are actually included. These are actually backported into https://github.com/systemd/systemd-stable v249.10, which is what we use for the debian package:
```
d48f9174cf211a235193963a06b3d28537fc6529 src/boot/efi/linux: fix linux_exec prototype
e98d271e57f3d0356e444b6ea2d48836ee2769b0 sd-boot: Rework console input handling
b52fafb26d90b77cfc259fcbdab3c95a571bacb1 sd-boot: Fix possible null pointer dereference
8aba0eec499b762657f528988c2f093ac490620d sd-boot: Unify error handling
2846007ecfb1fc84005b942167d394294c707d7b Fix the "Failed to open random seed ..." message.
```
`meson: use a/b instead of join_paths(a,b)` -> this one touched too many paths outside efi and was only cosmetic, so I decided to not include it.

`boot: stop making TPM PCR` -> This one breaks SB, see https://github.com/systemd/systemd/issues/22635

The ones not affecting files in `src/boot/efi/` are about moving around macros that are used in latest `src/boot/efi/`, so I had to include them for things to compile.

I have now created a PR for the non-upstream one, although not sure if it will be accepted: https://github.com/systemd/systemd/pull/22650

Thanks for going through the commits, I'll add a bit more info to the commit message so things are clearer.

valentindavid (approved):

alfonsosanchezbeato:
I have repushed, adding more explanations in the commit message. Also, [0073](https://github.com/systemd/systemd/pull/22650) has already been merged in systemd upstream.

alfonsosanchezbeato:
I have rebased to run the snapd smoke tests that are available now.

mvo5 (approved):
Given that this is cherry picks of upstream commits and very targeted to boot/efi I think we should merge it so that we have it for UC22 beta. *If* it turns out to be cause issues this will give us time to revert again.

alfonsosanchezbeato:
Thanks for the reviews, merging - agreed with @mvo5 that we have time to revert if it is problematic in the end.

PR #87, opened 2022-03-02
spread: refactor pipelines and add snapd testing

Meulengracht:
Refactor the test pipeline to seperate build and test stages. 
Include support for running spread tests from snapd (smoke-tests so far)

I will be doing some further fixes, I am initially creating this PR to test the github actions


tests/lib/prepare-utils.sh
```diff
@@ -0,0 +1,314 @@
...
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/snapd/snapd.spread-tests-run-mode-tweaks.sh
+RemainAfterExit=true
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+cat > "$snapddir/usr/lib/snapd/snapd.spread-tests-run-mode-tweaks.sh" <<'EOF'
```
valentindavid, line 201: Should not this be indented?

Meulengracht: Really good catch, it should be indented!

alfonsosanchezbeato (approved):
LGTM, thanks

valentindavid (approved):

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,129 @@
...
+    runs-on: self-hosted
+    steps:
+      - name: Cleanup job workspace
+        id: cleanup-job-workspace
+        run: |
+          rm -rf "${{ github.workspace }}"
+          mkdir "${{ github.workspace }}"
+      - uses: actions/checkout@v2
+
+      - name: Build snap
```
valentindavid, line 24: s/snap/deb/

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,129 @@
...
+        run: |
+          spread -artifacts=./artifacts google-nested:tests/spread/build/
+          find ./artifacts -type f -name "*.deb" | head -n 1 | while read line ; do cp "$line" "${{ github.workspace }}/uc-initramfs.deb" ; done
+      
+      - uses: actions/upload-artifact@v2
+        with:
+          name: initramfsdeb
+          path: "${{ github.workspace }}/uc-initramfs.deb"
```
valentindavid, line 32: It is not really safe in general to build a package in 22, and then install on 20. In this case, the dependencies for the current branch might not be available on ubuntu 20. So a safer thing to do would be to export the pc-kernel snap with initrd inject.

.github/workflows/tests.yaml
valentindavid: We should make a snap instead anyway.

PR #91, opened 2022-03-24
Add splash section to kernel.efi

xnox:
So I was hoping that this will give splash upon kernel.efi boot.

but it does not seem to work for me.

Either systemd-boot stub is broken, or my system doesn't support EFI graphics, or the image is somehow broken.

It would be nice to investigate this to make it work on real hardware. Or figure out if systemd boot stub splash support should be removed altogether.


valentindavid:
I have tested on NUC. I have added "quiet" and "splash" in the command line (though the second one is probably useless in this case). The splash showed for a very short time. The colors were wrong. It looked kind of green.

vlcsnap-2022-03-30-15h53m20s213

xnox:
> I have replaced locally the image with a 24 bits RGB BMP and the colors are good. It also worked for me on qemu. So if you could replace the BMP with a 24bits file, then I will approve the PR.

Given how small it looked, I think I better request proper graphic from design team.

The green stuff was probably transperancy?! Cause I've tried to have the logo have transparent things on the left and right, but maybe that is not needed at all.

xnox:
Note that no new kernel cmdline args are needed, as in the logo display is not sensistive to neither `quiet` nor `splash`. With updated logo I can now see it in qemu too.

However, when booting with `quiet` specified, it does seem that something somewhere clears it quite early..... it would be nice if it would have stayed for longer. I wonder if systemd or getty startup in the initrd or running system and clear the screen straight away.

But imho this is still better, as there are frequent requests for kernel.efi to provide some sort of logo as soon as practically possible.

xnox:
core22-splash

Splash view from qemu VM with updated logo.

valentindavid (approved):

xnox:
Note that on larger displays (higher dpi / hd / 4k) the logo will appear smaller and more smart. As the logo is fixed sized, rather than dpi aware.

xnox:
> Hey, this is cool but we would need some input from the design team. Note also that there was some previous work involving starting plymouth from the initramfs (see https://github.com/snapcore/core-initrd/tree/alfonsosanchezbeato/plymouth-wip), and also that kernel's efifb driver would take the same image as the one displayed by EFI fw, so there might be some discontinuity depending on what we want (EFI presents splash from OEM, then we get the display from sd-stub, then kernel might take same splash as EFI, then plymouth...).
> 
> So, I think we need some better idea of what each piece should do and of the full design before merging. There is also the question of how customizable by vendors this should be, etc.

On Classic, we go from firware displaying OEM logo, followed by plymouth continuing to show OEM logo on top, and showing Ubuntu logo on the bottom. Which then transitions seamlessly into gdm login screen.

I think showing some logo; which is more than/different from OEM logo is needed. I have also been pondering if the --splash option should also vendor the logo into initramfs to show it using something else too (i.e. plymouth, or psplash).

This did not have input from design team. Previously, from many years ago, we had spinning ubuntu logo on Ubuntu Touch, and huge "(cof) Core" on like UC16.

I wish we somehow had lighter Plymouth. Also note, if we do add plymouth in the initrd, it may need integration with snapd recovery passwords input & recovery keys, as I guess both would be fighting to control the keyboard input.

Separately pondering if systemd-boot stub can somehow influence the kernel's efifb displayed logo.

alfonsosanchezbeato:
Closing as we use now plymouth for splash.

PR #93, opened 2022-04-01
Switch from lz4 to zstd compression

xnox:
UC22 kernels support zstd compressed initramfs, and after extensive
analysis the default compression on classic was changed to zstd -1 -T0
(see ubuntu-devel threads and analysis from waveform). This
compression method improves compression ratio, compression speed, as
well as reduces decompression memory required.

Further improvements can be achieved later with firmware and module
compression.


alfonsosanchezbeato (approved):
LGTM, thanks

valentindavid (approved):

Meulengracht (approved):
Really nice, LGTM!

PR #94, opened 2022-04-07
Fix overriding of ExecStart for recovery-chooser-trigger

valentindavid:
Because ExecStart can be added multiple times, when overriding we need
to reset it. Otherwise it is started first without override, and since
it does not start with `@`, it gets killed during switch.


valentindavid:
Before the fix, you can see that `snapd.recovery-chooser-trigger.service` has been restarted in main boot. We do not see the run from initramfs because it has not survived it.

 recovery-chooser-before

Here is the run after the fix.

recovery-chooser-after

PR #95, opened 2022-04-22
ubuntu-core-initramfs: make paths relative when using the --root

alfonsosanchezbeato:
os.path.join() throws away previous components when it finds one that
is an absolute path - we do not want that when using --root as many
defaults are absolute, so we use this function to make second and
later paths relative when --root option is specified


bin/ubuntu-core-initramfs
```diff
@@ -9,14 +9,25 @@ import glob
...
 
 
+# os.path.join() throws away previous components when it finds one
+# that is an absolute path - we do not want that when using --root as
+# many defaults are absolute, so we use this function to make second
+# and later paths relative.
+def path_join_make_rel_paths(path, *paths):
+    relative_paths = []
+    for p in paths:
+        relative_paths.append(p[1:] if len(p) > 0 and p[0] == '/' else p)
```
valentindavid, line 19: What happens when the path is `//foo`?
Maybe it is better to use `os.path.relpath(p, '/')`

alfonsosanchezbeato: Good point, changed now

bin/ubuntu-core-initramfs
```diff
@@ -9,14 +9,25 @@ import glob
...
 
 
+# os.path.join() throws away previous components when it finds one
+# that is an absolute path - we do not want that when using --root as
+# many defaults are absolute, so we use this function to make second
+# and later paths relative.
+def path_join_make_rel_paths(path, *paths):
+    relative_paths = []
+    for p in paths:
+        relative_paths.append(os.path.relpath(p, '/') if len(p) > 0 and p[0] == '/' else p)
```
valentindavid, line 19: `os.path.relpath(p, '/') if os.path.isabs(p) else p`

or

 `os.path.relpath(os.path.join('/', p), '/')`

alfonsosanchezbeato: Changed now, thanks!

valentindavid (approved):

Meulengracht (approved):
LGTM!

PR #97, opened 2022-04-28
Fix recovery chooser trigger on NUC

valentindavid:
We need to chroot when switching root, so we send SIGUSR1 to recovery
chooser trigger.

We need to load all hid drivers. We should not load them into the
kernel though.

This works with https://github.com/snapcore/snapd/pull/11742

### DIFF
--- a/factory/usr/lib/modules-load.d/ubuntu-core-initramfs.conf
+++ b/factory/usr/lib/modules-load.d/ubuntu-core-initramfs.conf
@@ -2,7 +2,7 @@
 i2c-bcm2708
 nvme
 usbhid
-hid-generic
+xhci-pci
 squashfs
 ahci
 libahci
--- a/factory/usr/lib/recovery-chooser-trigger-switch-root
+++ b/factory/usr/lib/recovery-chooser-trigger-switch-root
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+set -eux
+
+kill -USR1 $(systemctl show --property MainPID --value snapd.recovery-chooser-trigger.service)
--- a/factory/usr/lib/systemd/system/initrd-switch-root.target.wants/recovery-chooser-trigger-switchroot.service
+++ b/factory/usr/lib/systemd/system/initrd-switch-root.target.wants/recovery-chooser-trigger-switchroot.service
@@ -1,0 +1,1 @@
+../recovery-chooser-trigger-switchroot.service
--- a/factory/usr/lib/systemd/system/recovery-chooser-trigger-switchroot.service
+++ b/factory/usr/lib/systemd/system/recovery-chooser-trigger-switchroot.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=Recovery chooser trigger switch root service
+DefaultDependencies=no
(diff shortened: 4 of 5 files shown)


factory/usr/lib/modules-load.d/ubuntu-core-initramfs.conf
```diff
@@ -2,7 +2,7 @@
 i2c-bcm2708
 nvme
 usbhid
-hid-generic
+xhci-pci
```
Meulengracht, line 5: can you explain this change? does this load all usb drivers?

valentindavid: The removed line is because now we have all the hid modules in initramfs because of the change in `bin/ubuntu-core-initramfs` (see call to `dracut-install ... =drivers/hid`).

The new line is to add support for USB 3.x controllers which is what most computers have now.

valentindavid: Note that `modules-load.d` is meant to force load modules in the kernel. But we do not need to do that.

Somewhere else, those configuration files are scanned to add the modules into initramfs. But we should only add modules that will be force loaded.

bin/ubuntu-core-initramfs
```diff
@@ -73,6 +73,22 @@ def create_initrd(parser, args):
...
+                "-D",
+                main,
+                "-r", kernel_root,
+                "--kerneldir",
+                modules,
+                "--firmwaredirs",
+                firmware,
+                "--module",
+                "--optional",
+                "=drivers/hid",
```
alfonsosanchezbeato, line 88: How much bigger is the initramfs after adding this modules?

factory/usr/lib/recovery-chooser-trigger-switch-root
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -eux
+
+kill -USR1 $(systemctl show --property MainPID --value snapd.recovery-chooser-trigger.service)
```
alfonsosanchezbeato, line 5: I'm +1 on using a signal to notify the pivot root.

alfonsosanchezbeato (commented):
Looks good, just have a question below.

factory/usr/lib/modules-load.d/ubuntu-core-initramfs.conf
Meulengracht: It makes sense to of course add support for USB 3.X, very nice change - this doesn't affect support for boards with only USB 2.0 controllers right?

valentindavid: That should not.

bin/ubuntu-core-initramfs
valentindavid: Approximately 1.4M more for x86_64 (from ~27M to ~28M).

alfonsosanchezbeato:
This can leverage the change to include modules without adding to modules.d that has been just merged.

PR #100, opened 2022-05-24
many: add folder with files specifying kernel modules for initramfs

alfonsosanchezbeato:
Previously, the only way to include kernel modules in the initramfs
was to add them to files in factory/usr/lib/modules-load.d/, which
forces the modules to be loaded. Use now per-feature configuration
files so we can include the modules without forcing their loading.

Use this feature to include the dwc2 driver for linux-raspi kernel.
On 5.15 kernel (jammy), this module needs to be included in the
initramfs to have proper USB support at that stage. One side effect of
not having this was that the ethernet driver got loaded later than in
previous releases.

### DIFF
--- a/bin/ubuntu-core-initramfs
+++ b/bin/ubuntu-core-initramfs
@@ -20,6 +20,31 @@ def path_join_make_rel_paths(path, *paths):
     return os.path.join(path, *relative_paths)
 
 
+# Include kernel modules specified in conf_file
+def add_modules_from_file(dest_d, kernel_root, modules_d, fw_d, conf_file):
+    with open(conf_file) as f:
+        for module in f.readlines():
+            module = module.strip()
+            if module:
+                if module[0] in ["#", ";"]:
+                    continue
+                subprocess.check_call(
+                    [
+                        "/usr/lib/dracut/dracut-install",
+                        "-D",
+                        dest_d,
+                        "-r", kernel_root,
+                        "--kerneldir",
+                        modules_d,
+                        "--firmwaredirs",
+                        fw_d,
+                        "--module",
+                        "--optional",
+                        module,
+                    ]
+                )
+
+
 def create_initrd(parser, args):
     if not args.kerneldir:
         args.kerneldir = "/lib/modules/%s" % args.kernelver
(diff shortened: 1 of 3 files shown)


valentindavid:
I think we should put the file in initramfs, But force loading it might not be right. If we just add it to the initramfs, it should be automatically loaded if the hardware is present.

In #97 we need to add lots of hid drivers. And I did not use `modules-load.d` for that reason. Maybe we should create a new configuration file that is not in `modules-load.d` but is still read by `bin/ubuntu-core-initramfs` to add the need modules.

xnox:
> I think we should put the file in initramfs, But force loading it might not be right. If we just add it to the initramfs, it should be automatically loaded if the hardware is present.
> 

For drivers that support that. We are still kind of forced to load virtio_scsi on all systems, because it is not otherwise auto-triggered.

> In #97 we need to add lots of hid drivers. And I did not use `modules-load.d` for that reason. Maybe we should create a new configuration file that is not in `modules-load.d` but is still read by `bin/ubuntu-core-initramfs` to add the need modules.

It is correct that currently the only way for any feature to include kernel module is to specify them to be included & always loaded via `modules-load.d` snippets. I wonder if we need a custom `modules-install.d` or some such? Same format as `modules-load.d` kernel modules and their firmware copied, but nothing reads that dir. It should probably have README as well, to explain what it is, given it will look unused/pointless "is it a typo for modules-load.d?"

alfonsosanchezbeato:
I've re-targeted this PR to now create a way to include additional kernel modules without forcing their loading from initramfs.

factory/usr/lib/modules-load.d/ubuntu-core-initramfs.conf
```diff
@@ -36,4 +36,5 @@ genet
 ## in-theory boot on raspi with the right dtb, so just keep it in by
 ## default
 vc4
-
+# raspi USB OTG driver, needed for jammy
+dwc2
```
valentindavid, line 40: Is this still needed here?

valentindavid:
Good. Thank you.

Tested building the initramfs and made sure them module was there.

valentindavid (approved):

alfonsosanchezbeato:
@xnox could you please take another look at this one?

Meulengracht (approved):
LGTM. Thanks!

PR #101, opened 2022-06-09
add required drivers on Intel platform

jianhuilee:


xnox (approved):

alfonsosanchezbeato:
Do all these modules need to be force-loaded? If not, maybe it will be better to wait for https://github.com/snapcore/core-initrd/pull/100 to be merged, so they could be included in the initramfs without forcing the loading.

valentindavid:
#97 already adds  xhci-pci. Could we get it merged instead.

jianhuilee:
Revised the patch based on #100 and #97 as suggested.

jianhuilee:
@alfonsosanchezbeato may i ask your help to review this? thanks.

valentindavid (approved):

alfonsosanchezbeato (approved):
LGTM

valentindavid:
xhci-pci-renesas is a dependency of xhci-pci so it was useless to add it. We should to remove it.

PR #102, opened 2022-06-15
Support for splash screen

alfonsosanchezbeato:
Adds configuration to the initramfs for splash screen. This requires also base changes. Just a draft for the moment as we need a new plymouth deb package in jammy and some snapd changes.


alfonsosanchezbeato:
This is ready for an initial pass now. Note that to have the full feature the changes in the base are needed, and a gadget with additional `quiet splash` kernel command line parameters is needed.

factory/usr/lib/systemd/system/plymouth-start.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Show Plymouth Boot Screen
+DefaultDependencies=no
+Wants=systemd-ask-password-plymouth.path systemd-vconsole-setup.service
+After=systemd-vconsole-setup.service systemd-udev-trigger.service systemd-udevd.service
+After=snap-initramfs-mounts.service
```
valentindavid, line 6: This sounds quite late.
 - We should have a splash before disks are scanned.
 - If `snap-bootstrap` needs a password, it will call `systemd-ask-password`, which if plymouth is properly started, will use `systemd-ask-password-plymouth.{path,service}` (units we need in initramfs), to ask the password within plymouth.

If we need to load a theme from the gadget, I would suggest we switch theme after `snap-bootstrap`.

alfonsosanchezbeato: As discussed, we will do this later when have the patch to switch themes in the jammy deb.

alfonsosanchezbeato:
Pushed now changes to use deb definition of the services and do overrides for the changes we need.

factory/usr/lib/systemd/system/plymouth-start.service
valentindavid: We should still put `systemd-ask-password-plymouth.{path,service}` in the initramfs.

factory/usr/lib/systemd/system/plymouth-start.service.d/core-override.conf
```diff
@@ -0,0 +1,6 @@
+[Unit]
+After=snap-initramfs-mounts.service
+
+[Service]
+ExecStart=
+ExecStart=/usr/sbin/plymouthd --debug --mode=boot --pid-file=/run/plymouth/pid --attach-to-session --ignore-serial-consoles
```
valentindavid, line 6: I suppose we can remove `--debug`.

alfonsosanchezbeato: Done

alfonsosanchezbeato (commented):
Thanks! Comment addressed, I've also included now `systemd-ask-password-plymouth.{service,path}` to be ready for future password handling.

debian/rules
```diff
@@ -213,6 +213,18 @@ override_dh_auto_install:
...
+
+	# splash functionality
+	mkdir -p debian/tmp/usr/share/plymouth/themes/
+	cp -a vendor/plymouth-theme-ubuntu-core/ubuntu-core \
+		debian/tmp/usr/share/plymouth/themes/
+	mkdir -p debian/tmp/usr/share/fonts/
+	cp /usr/share/fonts/truetype/ubuntu/Ubuntu-R.ttf \
+		debian/tmp/usr/share/fonts/Plymouth.ttf
+	# We need the seat rules for drm
+	sed 's#{{ROOTBINDIR}}#/bin#' vendor/systemd/src/login/71-seat.rules.in > \
```
valentindavid, line 225: This is incomplete. I get errors printed for lines 59 and 63.

alfonsosanchezbeato: Nice catch, I've fixed now

valentindavid:
There is a cycle issue.

`vendor/systemd/units/systemd-ask-password-console.path` has `After=plymouth-start.service`. And `Before=cryptsetup.target` which means `Before=sysinit.target`

If we add `After=snap-initramfs-mounts.service` to `plymouth-start.service`, then it means `snap-initramfs-mounts.service` cannot be `After=sysinit.target`.

alfonsosanchezbeato:
> There is a cycle issue.
> 
> `vendor/systemd/units/systemd-ask-password-console.path` has `After=plymouth-start.service`. And `Before=cryptsetup.target` which means `Before=sysinit.target`
> 
> If we add `After=snap-initramfs-mounts.service` to `plymouth-start.service`, then it means `snap-initramfs-mounts.service` cannot be `After=sysinit.target`.

I've removed the systemd-ask-password units now. Once we have the plymouth patch to switch themes landed we will start plymouth before snap-initramfs-mount, as we won't need the gadget snap mounted.

debian/rules
```diff
@@ -213,6 +213,19 @@ override_dh_auto_install:
...
+
+	# splash functionality
+	mkdir -p debian/tmp/usr/share/plymouth/themes/
+	cp -a vendor/plymouth-theme-ubuntu-core/ubuntu-core \
+		debian/tmp/usr/share/plymouth/themes/
+	mkdir -p debian/tmp/usr/share/fonts/
+	cp /usr/share/fonts/truetype/ubuntu/Ubuntu-R.ttf \
+		debian/tmp/usr/share/fonts/Plymouth.ttf
+	# We need the seat rules for drm. Replace jinja template bits for that.
+	sed -e '/{% raw [ -]%}/d' -e '/{% endraw %}/d' -e 's#{{ROOTBINDIR}}#/bin#' \
```
valentindavid, line 225: I do think we should do that. Could we instead enable logind and filter the installed binaries?

alfonsosanchezbeato: ok, this is done now

valentindavid (approved):
Great!

Meulengracht (approved):

PR #104, opened 2022-06-17
Print warnings when modules are in the wrong configuration file

valentindavid:
Modules with aliases are usually automatically loaded and do not need
to be in `modules-load.d`. Instead they should be included in the
initramfs by adding them in `extra-modules.conf`. Modules with symbol
aliases modules do not even need to be in `extra-modules.conf` since
`dracut-install` will find them if they are needed.

We also check the name against some known patterns. For example NLS
modules do not have alias, but are still loaded automatically.

Those which have aliases are moved from `modules-load.d` to
`extra-modules.conf`.
- `aes`
- `aesni-intel`
- `ahci`
- `autofs4`
- `cbc`
- `dm-crypt`
- `dm-mod`
- `echainiv`
- `fixed`
- `genet`
- `gpio-regulator`
- `hid-generic`
- `hv_storvsc`
- `hv_vmbus`
- `megaraid_sas`
- `mmc-block`
- `mptspi`
- `nvme`
- `phy-generic`
- `sdhci-acpi`
- `sdhci-pci`
- `sha256`
- `squahsfs`
- `usb-storage`
- `usbhid`
- `virtio_blk`
- `virtio_scsi`
- `xts`

Some do not have alias, but have a known automatic prefix:
- `algif-sckipher`
- `nls_iso8859-1`
- `dm-crypt`
- `dm-mod`

Those that have symbol aliases should be removed
completely. `dracut-install` can find them.
- `scsi_transport_sas`
- `libahci`
- `sdhci`

There are some unknown modules for x86_64. Those should be moved in
arch specific files. This commit does not fix this yet.
- `i2c-bcm2708`
- `sdhci-iproc`
- `vc4`


valentindavid:
I need to refactor and cleanup the script a bit. I will make it as ready when it looks better.

bin/ubuntu-core-initramfs
```diff
@@ -21,13 +115,32 @@ def path_join_make_rel_paths(path, *paths):
 
 
 # Include kernel modules specified in conf_file
-def add_modules_from_file(dest_d, kernel_root, modules_d, fw_d, conf_file):
+def add_modules_from_file(dest_d, kernel_root, modules_d, fw_d, conf_file, db,
+                          allow_discoverable=True):
```
alfonsosanchezbeato, line 119: Some description of the function parameters (old&new) would be nice to have here.

alfonsosanchezbeato (commented):
Thanks for this! Some notes:
- I'm not sure if the modules that do not have a hw info alias are loaded. I don't know what would be the mechanism in this case. In a quick test on my laptop, `sha256` needs to be manually loaded, so I think we should still force-load them unless we know how they would get loaded otherwise.
- Regarding the ones that have symbol aliases, I assume those would be pulled by dracut only in case there are modules included that require them as dependencies. Could you please point out the modules that depend on the 3 modules that have symbol aliases?

bin/ubuntu-core-initramfs
```diff
@@ -21,13 +115,32 @@ def path_join_make_rel_paths(path, *paths):
...
+                    print(f"NOTE: {conf_file}: Module {module} not found", file=sys.stderr)
+                else:
+                    if mod.builtin:
+                        print(f"NOTE: {conf_file}: Module {module} is builtin", file=sys.stderr)
+                    if not allow_discoverable and len(mod.aliases) > 0:
+                        print(f"WARNING: {conf_file}: Module {module} has aliases:", file=sys.stderr)
+                        for alias in mod.aliases:
+                            print(f" * {alias}", file=sys.stderr)
+                    # A module that has symbols but no alias is likely to be loaded through aliases
+                    if len(mod.aliases) == 0 and len(mod.symbols) > 0:
```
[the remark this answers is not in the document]
valentindavid: I have pushed a commit adding this functionality. But it does not find more. `len(mod.aliases) == 0 and len(mod.symbols) > 0` is better because it also finds modules that are truly useless. If it is not installed by other modules, using dependencies would not find it. So I think it is redundant, and I am planning to remove that commit. Tell me what you think.

bin/ubuntu-core-initramfs
valentindavid: Done

valentindavid:
I have found that `modprobe` was missing. And this explain why many modules had problems to be automatically loaded. I added it, and now everything works as it should.

debian/rules
```diff
@@ -218,11 +218,28 @@ override_dh_auto_install:
...
+	for f in								\
+		/lib/$(DEB_HOST_MULTIARCH)/libnss_files.so.*			\
+		/lib/$(DEB_HOST_MULTIARCH)/libnss_compat.so.*			\
+		/lib/$(DEB_HOST_MULTIARCH)/libgcc_s.so.1			\
+		/sbin/e2fsck							\
+		/sbin/fsck.vfat							\
+		/sbin/fsck /bin/umount						\
+		/bin/mount							\
+		/bin/kmod							\
+		/usr/sbin/modprobe						\
```
alfonsosanchezbeato, line 231: Is this preserving modprobe as a symlink or does it make a copy of kmod? Also, I think this is a good opportunity to add the rest of possible symlinks to kmod: lsmod, rmmod, insmod, modinfo, depmod. It is always annoying having to create them by hand when debugging the initramfs.

alfonsosanchezbeato (commented):
Looks good, just see comment below. Also, the commit message in 462e36f38 mentions that all modules that have aliases are moved, perhaps it would be more accurate to say that the ones that have hw aliases are loaded when the hw is detected so they can be moved, and that others that don't have hw aliases are automatically loaded by the kernel when some syscalls need them by using modprobe (as aes) - having aliases or not for these is irrelevant afaiu. The latest is true as well for the ones with known automatic prefixes.

debian/rules
valentindavid: `dracut-install` preserves the symlinks. I have added the missing ones.

alfonsosanchezbeato:
> > Also, the commit message in [462e36f](https://github.com/snapcore/core-initrd/commit/462e36f38e841118a44a28bfa2aa21ba6cf731b0) mentions that all modules that have aliases are moved, perhaps it would be more accurate to say that the ones that have hw aliases are loaded when the hw is detected so they can be moved, and that others that don't have hw aliases are automatically loaded by the kernel when some syscalls need them by using modprobe (as aes) - having aliases or not for these is irrelevant afaiu. The latest is true as well for the ones with known automatic prefixes.
> 
> Aliases are mainly made for the kernel to be able to automatically load modules. That they are detected through a hardware bus or during a syscall, it is still just the kernel calling `request_module`. So an alias is a strong indication that the module knows how to load it. So there is no real difference between hardware and syscalls.

Well, there is a difference. For instance, in these two lines:

```
alias aes aesni_intel
alias cpu:type:x86,ven*fam*mod*:feature:*0099* aesni_intel
```

The second line makes sure that `aesni_intel` is loaded by udev when the hardware is detected. The first line is used by the kernel when something that needs AES is run, to check if some already loaded module provides an implementation, moving otherwise to using software implementation. If the only line was

```
alias aes aesni_intel
```

That would not load automatically `aesni_intel`. The syscall would actually load `aes.ko` because is sort of hard-coded in the code. Anyway, it is possibly true that most of what is in modules.alias is automatically loaded in one or other way.

> 
> > Finally, for completeness, what modules pull `scsi_transport_sas`, `libahci`, and `sdhci`?
> 
> Those are like libraries. Modules that share code for multiple modules. So it does not matter what uses them, they will always be loaded automatically. And if they are not, they are useless.
> 
> If you want the details
> 
>     * `scsi_transport_sas` is needed by `megaraid_sas`.
> 
>     * `libahci` is needed by `ahci`.
> 
>     * `sdhci` is needed by `sdhci-*`
> 
> 
> There might be other modules that need them.

valentindavid:
Just for precision, there is no `aes.ko`. There is `aes-generic.ko` which has an alias. Also `aes` is an old alias, probably kept for backward compatibility for systems that do not let the kernel automatically load modules. `crypto-aes` is the one currently used.

Meulengracht (approved):
LGTM, this looks nice!

PR #107, opened 2022-07-28
Wait for /dev/ubuntu/seed.device to start snap-initramfs-mounts.service

valentindavid:


valentindavid:
This requires snapcore/snapd#12006

xnox:
It can often be confusing if udev database contents is different between initrd and runtime system. I hope that this udev rule will be added to the core* snap, and snapd.deb as well. Such that Links= on the system for the given partitions are the same in initrd and running systems.

xnox:
I forgot what this is for anymore.

valentindavid:
> I forgot what this is for anymore.

snap-bootstrap starts before the disk is detected, which means it needs to loop and sleep until the disk appears. This is not elegant nor efficient. This PR changes it to react to udev and starts snap-bootstrap once the disk is detected without any sleep.

PR #108, opened 2022-08-05
releasing package ubuntu-core-initramfs version 56

alfonsosanchezbeato:


debian/changelog
```diff
@@ -1,3 +1,20 @@
...
+  * Fix recovery chooser trigger on NUC
+  * Print warnings when modules are in the wrong configuration file
+  * debian/rules: Remove more unused systemd files
+  * Mount UC mounts only when UC is detected
+
+  [ Jian Hui Lee ]
+  * add required drivers on Intel platform
+
+  [ Alfonso Sanchez-Beato ]
+  * many: add folder with files specifying kernel modules for initramfs
```
valentindavid, line 14: I think you also added a module for rpi.

debian/changelog
```diff
@@ -1,3 +1,20 @@
+ubuntu-core-initramfs (56) jammy; urgency=medium
+
+  [ Valentin David ]
+  * Update systemd to 249.11-0ubuntu3
+  * Fix recovery chooser trigger on NUC
+  * Print warnings when modules are in the wrong configuration file
```
valentindavid, line 6: I think for the perspective of the user we should say something like "Do not force load kernel modules that can be automatically loaded".

valentindavid: or "explicitly load"

valentindavid (approved):

PR #109, opened 2022-08-11
Forward-port CVM

xnox:
Given all the units refactoring, and starts of classic mode, majority of cloudimg-rootfs feature now can simply be deleted.

A custom snap-bootstrap is still required (which has https://github.com/snapcore/snapd/pull/10267), plus just condition tweaks on the kernel mount (CVM does not use kernel snap, hence no need to mount/bind-mount kernel snap modules and firmware)

This works on jammy for me.


alfonsosanchezbeato (approved):
LGTM

PR #110, opened 2022-08-15
releasing package ubuntu-core-initramfs version 57

xnox:
Upload ubuntu-core-initramfs with cloudimg-rootfs feature fixedup


alfonsosanchezbeato (approved):
LGTM

PR #111, opened 2022-08-23
Do not forward journal to console when quiet

valentindavid:
Also:

Parse /proc/cmdline correctly
    
There is potential quotes in kernel command line and we need to handle those correctly.
    
For more information on how to parse kernel parameters, see function `next_arg` in https://github.com/torvalds/linux/blob/master/lib/cmdline.c

### DIFF
--- a/factory/usr/libexec/core/get-arg
+++ b/factory/usr/libexec/core/get-arg
@@ -1,0 +1,84 @@
+#!/bin/sh
+
+# Usage:
+#   get-arg param-name
+#
+# get-arg will look for kernel parameter "param-name" or "param_name"
+# and return 0 if found, 1 if not found. If the parameter as a value,
+# e.g. "param-name=the-value", then the value "the-value" will be
+# printed.
+
+# For more information on how to parse kernel parameters, see function
+# `next_arg` in
+# https://github.com/torvalds/linux/blob/master/lib/cmdline.c
+
+set -eu
+
+if [ "$#" -ne 1 ]; then
+    echo "Expected kernel parameter name as argument" 1>&2
+    exit 1
+fi
+
+looking_for="$(echo "${1}" | sed 's/_/-/g')"
+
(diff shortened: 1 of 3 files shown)


factory/usr/libexec/core/get-arg
```diff
@@ -0,0 +1,79 @@
...
+            ;;
+    esac
+done
+
+if [ -n "${param}" ]; then
+    set -- "$@" "${param}"
+fi
+
+for param in "$@"; do
+    name="$(echo "${param%=*}" | sed 's/_/-/g')"
```
alfonsosanchezbeato, line 68: Should this be `%%=`?. First equal to separate name from argument with some = around is how shell would do it, but I do not know the convention for the kernel.

factory/usr/libexec/core/get-arg
```diff
@@ -0,0 +1,79 @@
...
+    if [ "${name}" = "${looking_for}" ]; then
+        case "${param}" in
+            *=*)
+                echo "${param##*=}"
```
alfonsosanchezbeato, line 72: And `#*=` here?

alfonsosanchezbeato (commented):
Please see question, but looks good otherwise

factory/usr/libexec/core/get-arg
valentindavid: Correct. We want the longest match.

factory/usr/libexec/core/get-arg
valentindavid: Correct. We want the shortest match.

alfonsosanchezbeato (approved):
Thanks for the changes. Nitpick: maybe it would be good to return with error if `get-arg` is called without an argument or with the wrong number of arguments.

valentindavid:
> Thanks for the changes. Nitpick: maybe it would be good to return with error if get-arg is called without an argument or with the wrong number of arguments.

Done.

Meulengracht (approved):
LGTM

PR #114, opened 2022-09-08
vendor/systemd: maintain stable default net-naming-scheme

xnox:
An upcoming systemd sru will introduce additional (future) naming schemes, to ensure backwards compatibility explicitly configure systemd networkd udevd net-naming-scheme to remain the same, at v245, as before.

See related LP & focal classic SRU.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -72,6 +72,7 @@ override_dh_auto_configure:
       -Dwheel-group=false \
       -Ddev-kvm-mode=0660 \
       -Dgroup-render-mode=0660 \
+      -Ddefault-net-naming-scheme=v245 \
       -Ddefault-dnssec=no \
       -Dselinux=true \
       -Ddefault-kill-user-processes=false \


xnox:
`meson: error: unrecognized arguments:  # LP:`

great

alfonsosanchezbeato:
Thanks for this, I was not aware of this SRU. The patches from https://code.launchpad.net/~ubuntu-core-dev/ubuntu/+source/systemd/+git/systemd/+merge/409304 would need to be applied too though.

xnox:
The new systemd is already vendored and included in core20. so this should be merged as well and released. Or maybe should have been included as part of https://github.com/snapcore/core-initrd/pull/130 which this was missed.

alfonsosanchezbeato (approved):
LGTM (the patches needed were now merged)

alfonsosanchezbeato:
I've rebased to master now.

PR #116, opened 2022-09-09
modules: include intel drm in initramfs

alfonsosanchezbeato:
So we can have splash with intel GPUs.

### DIFF
--- a/modules/main/extra-modules.conf
+++ b/modules/main/extra-modules.conf
@@ -44,3 +44,5 @@ vmd
 xhci-pci-renesas
 # qemu drm (older kernels: bochs_drm)
 bochs
+# intel drm
+i915


valentindavid:
Should we also add amdgpu, nouveau, radeon, gm500-gfx and virtio-gpu?

alfonsosanchezbeato:
> Should we also add amdgpu, nouveau, radeon, gm500-gfx and virtio-gpu?

That's a good question... some of those are a bit on the big side (amdgpu is 20MB). I want to add the Intel one because we run UC on Intel NUCs, but I'm not sure of use cases for the others, although it would be nice to have them. I'd say that we add ones that we know for sure that have a use case - virtio-gpu probably. @mvo5 @kubiko what do you think?

valentindavid:
Note that AMD has equivalent boards called "Mini PC".

valentindavid:
amdgpu is need for integrated gpu on ryzen.

valentindavid (approved):

alfonsosanchezbeato:
I think that the criteria should be whether we have UC certified devices with such a GPU. We have certified intel NUCs, but I don't think that's the case for mini PCs or others. I think we can merge this and have follow-up PRs as needed.

PR #118, opened 2022-09-26
Fix unit dependencies

valentindavid:
Break cycle introduced in #102 with plymouth and systemd-password. It was discovered in https://github.com/snapcore/core-initrd/pull/102#issuecomment-1222297471 but was not properly fixed.

Also remove `populate-writable.service` from `basic.target`. It is not needed.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -241,6 +241,8 @@ override_dh_auto_install:
 		rm -r debian/tmp/$$p ;		\
 	done
 	touch debian/tmp/etc/machine-id
+	# This hack should be removed with #113
+	sed -i '/^After=/{;s, *plymouth-start[.]service *, ,;/^After= *$$/d;}' debian/tmp/usr/lib/systemd/system/systemd-ask-password-*
 
 	LD_LIBRARY_PATH=`pwd`/debian/tmp/usr/lib/*/:`pwd`/debian/tmp/usr/lib/systemd `pwd`/debian/tmp/usr/bin/systemd-hwdb --root debian/tmp update --usr --strict
 	rm -rf debian/tmp/usr/lib/udev/hwdb.d
--- a/factory/usr/lib/systemd/system/basic.target.wants/populate-writable.service
+++ b/factory/usr/lib/systemd/system/basic.target.wants/populate-writable.service
@@ -1,1 +1,0 @@
-../populate-writable.service


alfonsosanchezbeato:
Shouldn't we just remove all `systemd-ask-password-*` services? We don't really need/want them at the moment.

valentindavid:
We want them to ask if someone needs to enter a recovery key. I am not sure if anyone has ever used it on Ubuntu Core and enrolled recovery keys. But removing it potentially removes a feature that has been present for a long time. I suppose removing it would make systemd-ask-password fail after a timeout.

Meulengracht:
Thanks! It could be nice to ask Field about if they know anything about recovery keys, and whether we have customers using it. It's probably gonna affect normal users the most, who use ubuntu core to play around with or for their own projects.

Meulengracht (approved):

debian/rules
```diff
@@ -241,6 +241,8 @@ override_dh_auto_install:
 		rm -r debian/tmp/$$p ;		\
 	done
 	touch debian/tmp/etc/machine-id
+	# This hack should be removed with #113
+	sed -i '/^After=/{;s, *plymouth-start[.]service *, ,;/^After= *$$/d;}' debian/tmp/usr/lib/systemd/system/systemd-ask-password-*
```
alfonsosanchezbeato, line 245: FTR this creates an empty `After=` in `systemd-ask-password-console.path`, which is probably fine

alfonsosanchezbeato (approved):
LGTM. Re: recovery keys, I'm curios, where are we actually handling input for that? I cannot see references to `systemd-ask-password`.

debian/rules
valentindavid: I do not see it.

```
cat debian/tmp/usr/lib/systemd/system/systemd-ask-password-console.path | grep After=
```

Gives nothing.

valentindavid:
> Re: recovery keys, I'm curios, where are we actually handling input for that? I cannot see references to systemd-ask-password.

secboot calls `systemd-ask-password`.

valentindavid:
https://github.com/snapcore/secboot/blob/master/auth_requestor_systemd.go#L48

Also I have seen working multiple time when I reset my tpm.

debian/rules
alfonsosanchezbeato: This is what I saw when applying the command in the file:
```
$ sed '/^After=/{;s, *plymouth-start[.]service *, ,;/^After= *$$/d;}' usr/lib/ubuntu-core-initramfs/main/usr/lib/systemd/system/systemd-ask-password-console.path
#  SPDX-License-Identifier: LGPL-2.1-or-later
#
#  This file is part of systemd.
#
#  systemd is free software; you can redistribute it and/or modify it
#  under the terms of the GNU Lesser General Public License as published by
#  the Free Software Foundation; either version 2.1 of the License, or
#  (at your option) any later version.

[Unit]
Description=Dispatch Password Requests to Console Directory Watch
Documentation=man:systemd-ask-password-console.path(8)
DefaultDependencies=no
Conflicts=shutdown.target emergency.service
After= 
Before=paths.target shutdown.target cryptsetup.target
ConditionPathExists=!/run/plymouth/pid

[Path]
DirectoryNotEmpty=/run/systemd/ask-password
MakeDirectory=yes
```

alfonsosanchezbeato:
> https://github.com/snapcore/secboot/blob/master/auth_requestor_systemd.go#L48
> 
> Also I have seen working multiple time when I reset my tpm.

Thanks for the pointer!

debian/rules
valentindavid: But the `$$` will be changed to `$` by `make`.

valentindavid: Try running the same command through make.

alfonsosanchezbeato: Oh, right, didn't notice that. Thanks!

PR #120, opened 2022-10-19
Move mount units into /etc/fstab

valentindavid:
Unfortunately because /etc/fstab is still generated from populate-writable, we need to dynamically write the missing entries. They will be moved to core-base when we will use a static /etc/fstab.

For the moment adding the entries in /etc/fstab makes systemd be able to keep track or where they come from (that is fstab). So we can eventually add options so that the mount file gets re-generated with all information into the main boot.

populate-writable gets moved before initrd-root-fs so that systemd does parse /etc/fstab before the root switch.


valentindavid:
This hack will be removed by #74

factory/usr/lib/core/extra-paths
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -eu
+
+cat <<EOF >>/sysroot/etc/fstab
+/run/mnt/data /writable none bind,x-initrd.mount 0 0
```
alfonsosanchezbeato, line 6: Isn't this what `sysroot-writable.mount` mounts? Shouldn't you remove it as for the kernel ones?

valentindavid: `/etc/fstab` is bind mounted from `/run` by `handle-writable-paths`. So we have to run this script after `handle-writable-paths`. `handle-writable-paths` also needs `/writable` to be mounted. So that means this script depends on `sysroot-writable.mount`.

alfonsosanchezbeato:
I have tested this PR now jointly with https://github.com/snapcore/core-base/pull/84 and https://github.com/snapcore/core-base/pull/83 and these errors that happened on shutdown:

```
[FAILED] Failed unmounting /usr/lib/modules.
...
[FAILED] Failed unmounting /run/mnt/data.
```

do not happen anymore.

alfonsosanchezbeato (approved):
LGTM, thanks

PR #121, opened 2022-11-01
factory,debian: include plymouth classic splash

alfonsosanchezbeato:
Include files for plymouth classic splash, and make sure it gets selected in case we are starting a classic rootfs.


valentindavid (approved):
That works.

We have to remember that we will need a common intro splash for when we will start plymouth before snap-bootstrap (and update the splash after).

PR #124, opened 2022-11-10
Wait 10 seconds before emergency reboot

valentindavid:


factory/usr/lib/systemd/system/emergency-reboot.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Wait a bit before emergency reboot
+DefaultDependencies=no
+After=emergency.target
+ConditionKernelCommandLine=!dangerous
```
mvo5, line 5: Pardon my ignorance - so am I reading this right:
1. f we have a non-dangerous commandline this will result in a 10s delay of the boot (which is exactly what we want)?
2.  And for dangerous cmdline the change below in emergency.target.d means we get dropped into a shell? 
3. And in (1) the wait happens right after a failed unit so that the error is visible and not right before the reboot when things may have scrolled by?

I guess yes is the answer to all of them but I'm not familar enough to be 100% confident :)

mvo5 (commented):
Thanks for this PR, that is super nice and helps a lot! Some questions inline just to double check that I understand it correctly.

mvo5 (approved):
Thanks

Meulengracht (approved):
Really nice! LGTM

PR #125, opened 2022-11-10
Make snap-initramfs-mounts.service required

valentindavid:
Because we now depend on the existence of `/run/mnt/base` to decide what system we are booting, we need to fail the boot if partially failed `snap-initramfs-mounts.service`. Otherwise we get confusing error about trying to boot a classic system on a core system.


mvo5 (approved):
Thank you!

PR #126, opened 2022-11-18
factory: populate fstab for classic with modes

alfonsosanchezbeato:
So we can load firmware/modules coming from the kernel snap. Fixes a regression from https://github.com/snapcore/core-initrd/pull/120.


alfonsosanchezbeato:
Closing in favor of https://github.com/snapcore/core-initrd/pull/120

PR #127, opened 2022-11-18
Add generator for kernel mount points on classic

valentindavid:
This generator should also be run in classic root. That means should be later moved into a package and installed on classic, and core-initrd should copy from there.

### DIFF
--- a/factory/usr/lib/systemd/system-generators/kernel-snap-generator
+++ b/factory/usr/lib/systemd/system-generators/kernel-snap-generator
@@ -1,0 +1,57 @@
+#!/bin/sh
+
+set -eu
+
+: "${KERNEL_MNT_POINT:=/run/mnt/kernel}"
+
+if ! mountpoint -q "${KERNEL_MNT_POINT}"; then
+    exit 0
+fi
+
+# FIXME use SYSTEMD_IN_INITRD when using systemd 251+
+if [ -f /etc/initrd-release ]; then
+    sysroot=/sysroot
+    sysroot_unit=sysroot
+    target=initrd-fs.target
+else
+    sysroot=
+    sysroot_unit=
+    target=local-fs.target
+fi
+
+: "${OS_RELEASE:="${sysroot}/etc/os-release"}"
+
+# Note that /sysroot/etc/os-release is accessible when generators are
+# re-executed due to "daemon-reload" called by
+# initrd-parse-etc.service
+
+# We generate the kernel bind mount units only
+# if we are on Ubuntu Classic
+
+# Ubuntu Core will have those mounts declared in /etc/fstab, thus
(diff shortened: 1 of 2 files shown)


factory/usr/lib/systemd/system-generators/kernel-snap-generator
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+set -eu
+
+: ${KERNEL_MNT_POINT:=/run/mnt/kernel}
```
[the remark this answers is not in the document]
valentindavid: Fixed.

factory/usr/lib/systemd/system-generators/kernel-snap-generator
```diff
@@ -0,0 +1,41 @@
...
+
+if [ "${SYSTEMD_IN_INITRD}" = 1 ]; then
+    sysroot=/sysroot
+    target=initrd-fs.target
+else
+    sysroot=
+    target=local-fs.target
+fi
+
+: ${OS_RELEASE:="${sysroot}/etc/os-release"}
```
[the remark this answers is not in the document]
valentindavid: Fixed.

factory/usr/lib/systemd/system-generators/kernel-snap-generator
```diff
@@ -0,0 +1,47 @@
...
+    target=initrd-fs.target
+else
+    sysroot=
+    sysroot_unit=
+    target=local-fs.target
+fi
+
+: "${OS_RELEASE:="${sysroot}/etc/os-release"}"
+
+if ! grep -q "^ID=ubuntu$" "${OS_RELEASE}"; then
```
alfonsosanchezbeato, line 24: The file won't exist when this is run from initramfs, is that expected?

valentindavid: I have commented.

alfonsosanchezbeato (approved):
Thanks!

factory/usr/lib/systemd/system-generators/kernel-snap-generator
```diff
@@ -0,0 +1,57 @@
...
+
+# Note that /sysroot/etc/os-release is accessible when generators are
+# re-executed due to "daemon-reload" called by
+# initrd-parse-etc.service
+
+# We generate the kernel bind mount units only
+# if we are on Ubuntu Classic
+
+# Ubuntu Core will have those mounts declared in /etc/fstab, thus
+# using systemd-fstab-generator instead.
```
xnox, line 32: Which `/etc/fstab` do we mean here? the one in the initrd or the one in the installed system?

If need be, we can and should modify classic system's `/etc/fstab` such that required mounts are listed there, and then drop this.

PR #130, opened 2022-12-02
[core20] Update systemd to 245.4-4ubuntu3.19

valentindavid:


alfonsosanchezbeato (approved):
LGTM . I've checked against the deb package to make sure sources are the same, but the two extra patches for UC.

Maybe it is now a good time to add information about how to use git-tree in the HACKING file in main branch, before we forget again :)

PR #131, opened 2022-12-05
Build for core24

valentindavid:


xnox:
main is open for development, should we merge this? or should we attempt to just stage/copy relevant systemd files from the stock Ubuntu archive build?

Issue #134, opened 2022-12-06
overlap between extra-modules.conf and ubuntu-core-initramfs.conf


kubiko:
version: ubuntu core 22
Using combination of
- extra modules in `${skeleton}/modules/main/extra-modules.conf`
- configured modules to be loaded by systemd in `${skeleton}/main/usr/lib/modules-load.d/ubuntu-core-initramfs.conf`

If those two lists contain the same kernel module(s), it makes `ubuntu-core-initramfs create-initrd` fail with the following error:
```
Traceback (most recent call last):
  File "/home/ondrak/kernel-snap/parts/kernel/build/ubuntu-core-initramfs/usr/bin/ubuntu-core-initramfs", line 490, in <module>
    main()
  File "/home/ondrak/kernel-snap/parts/kernel/build/ubuntu-core-initramfs/usr/bin/ubuntu-core-initramfs", line 486, in main
    globals()[args.subcmd.replace("-", "_")](parser, args)
  File "/home/ondrak/kernel-snap/parts/kernel/build/ubuntu-core-initramfs/usr/bin/ubuntu-core-initramfs", line 310, in create_initrd
    add_modules_from_file(main, kernel_root, modules, firmware, module_load, db,
  File "/home/ondrak/kernel-snap/parts/kernel/build/ubuntu-core-initramfs/usr/bin/ubuntu-core-initramfs", line 251, in add_modules_from_file
    db.mark_installed(module, conf_file)
  File "/home/ondrak/kernel-snap/parts/kernel/build/ubuntu-core-initramfs/usr/bin/ubuntu-core-initramfs", line 135, in mark_installed
    elif old_mode == ModuleDb.IMPLICIT:
AttributeError: type object 'ModuleDb' has no attribute 'IMPLICIT'
```

alfonsosanchezbeato:
Fixed by https://github.com/snapcore/core-initrd/pull/135

kubiko:
thank you @alfonsosanchezbeato!

PR #135, opened 2022-12-06
bin/ubuntu-core-initramfs: fix ubuntu-core-initramfs crash

alfonsosanchezbeato:
When having a module both in extra-modules.conf and in a file in main/usr/lib/modules-load.d/, "ubuntu-core-initramfs create-initrd" crashed because of a wrong value check in the script. This commit fixes that.

Reported in https://github.com/snapcore/core-initrd/issues/134


xnox:
Some spelling issues in the commit message "was crashing becaus of a wrong value check in the script" can be "crashed because of ..." or "crashed due to ...". But that's minor oversight, maybe spellcheck is not turned on in git commit messages editor.

xnox (approved):

PR #139, opened 2023-01-27
tests: avoid interactions when using apt

alfonsosanchezbeato:
Lately CI was failing because of some problem with configuration of openssh-server.


mvo5 (approved):
Thank you!

PR #143, opened 2023-02-23
spread: disable apt upgrade of grub-efi-amd64-signed

valentindavid:
CI is failing when upgrading grub.

### DIFF
--- a/tests/lib/prepare-utils.sh
+++ b/tests/lib/prepare-utils.sh
@@ -122,11 +122,12 @@ install_core_initrd_deps() {
     # from when we build the debian package
     sudo add-apt-repository ppa:snappy-dev/image -y
     sudo DEBIAN_FRONTEND=noninteractive apt update -yqq
+    sudo apt-mark hold grub-efi-amd64-signed
     sudo DEBIAN_FRONTEND=noninteractive apt upgrade -yqq
 
     # these are already installed in the lxd image which speeds things up, but they
     # are missing in qemu and google images.
-    sudo DEBIAN_FRONTEND=noninteractive apt install initramfs-tools-core psmisc fdisk snapd mtools ovmf qemu-system-x86 sshpass whois openssh-server -yqq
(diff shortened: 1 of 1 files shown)


xnox:
It means the instance was not booted with an uptodate kernel.  Can the instance upgrade it's kernel and reboot?

Or do we just need to wait for a newer base image to be used?

valentindavid:
Not sure if we can reboot. It is a github builder.

alfonsosanchezbeato:
This runs on self-hosted runners, which are lxd containers iirc, so it is fine to hold grub. But I'm not sure why it is installed, it does not look like it is needed by the other packages that are installed, so maybe it is better to remove instead of holding.

tests/lib/prepare-utils.sh
```diff
@@ -122,11 +122,12 @@ install_core_initrd_deps() {
     # from when we build the debian package
     sudo add-apt-repository ppa:snappy-dev/image -y
     sudo DEBIAN_FRONTEND=noninteractive apt update -yqq
+    sudo apt-mark hold grub-efi-amd64-signed
     sudo DEBIAN_FRONTEND=noninteractive apt upgrade -yqq
 
     # these are already installed in the lxd image which speeds things up, but they
     # are missing in qemu and google images.
-    sudo DEBIAN_FRONTEND=noninteractive apt install initramfs-tools-core psmisc fdisk snapd mtools ovmf qemu-system-x86 sshpass whois openssh-server -yqq
+    sudo DEBIAN_FRONTEND=noninteractive apt install --no-install-recommends initramfs-tools-core psmisc fdisk snapd mtools ovmf qemu-system-x86 sshpass whois openssh-server -yqq
```
xnox, line 130: initramfs-tools-core is obsolete and unused by core-initrd.... why is it being installed?

alfonsosanchezbeato (approved):
LGTM, thanks

tests/lib/prepare-utils.sh
valentindavid: Opened https://github.com/snapcore/core-initrd/pull/152

PR #144, opened 2023-03-06
Add GPU and input virtio modules

alfonsosanchezbeato:
These modules are for qemu support, and are especially useful for arm64.


xnox (approved):

alfonsosanchezbeato:
I've tested now, works as expected (splash screen starts from initramfs after adding `quiet splash` to kernel command line).

Meulengracht (approved):

PR #145, opened 2023-03-17
hacking.md: fix paths and file names

sergio-costas:
Some file names are wrong or are old.


HACKING.md
```diff
@@ -79,13 +79,13 @@ and yq (needed for yaml manipulation), download the newest image and import it i
 ```
 sudo snap install lxd
 sudo snap install yq
-curl -o lxd-initrd-img.tar.gz https://storage.googleapis.com/snapd-spread-core/lxd/lxd-spread-initrd22-img.tar.gz
-lxc image import lxd-initrd-img.tar.gz --alias ucspread22
+curl -o lxd-initrd-img.tar.gz https://storage.googleapis.com/snapd-spread-core/lxd/lxd-spread-initrd-img.tar.gz
+lxc image import lxd-spread-initrd-img.tar.gz --alias ucspread22
```
Meulengracht, line 83: This seems wrong as curl saves the output as `lxd-initrd-img.tar.gz` if you look at the `-o`

alfonsosanchezbeato:
@Meulengracht I guess that if that is the path that works, then this is fine, but shouldn't actually the image have a 22 somewhere in the name? What would happen when we have core24 images?

Meulengracht:
> @Meulengracht I guess that if that is the path that works, then this is fine, but shouldn't actually the image have a 22 somewhere in the name? What would happen when we have core24 images?

The link should actually be correct, if it's not correct it *may* have expired. We may have to manually run the lxd-image workflow

sergio-costas:
If you go to https://storage.googleapis.com/snapd-spread-core/ you can see that there is only one file, lxd/lxd-spread-initrd-img.tar.gz, while there are two lxd-spread-core: 20 and 22. So maybe what should be fixed is the file name...

Meulengracht:
I just tested the link and it's working now

PR #153, opened 2023-05-02
Add dbus system bus

valentindavid:
D-Bus system bus is needed by `systemd-run --wait`.


xnox:
the inevitable.... note i'm not sure we can restart dbus, it means, that this dbus will likely continue running in the pivotted system? or is it being killed as part of pivot?

valentindavid:
In theory, it should be stopped when isolate to `initrd-switch-root.target` happen, since it is not a dependency of it. I will verify that it behaves properly.

valentindavid:
I have manually tested. D-Bus stops before switch root.

xnox (approved):

PR #156, opened 2023-05-04
spread: use older ubuntu-image

valentindavid:


Meulengracht (approved):
I mean sucks we need this but thanks! +1

alfonsosanchezbeato (approved):
LGTM

PR #157, opened 2023-05-10
core22: Add build-conflicts on snapd 2.59.3

xnox:
add build-conflicts on the snap-bootstrap version that has caused a recent regression. This is a safety precaution to prevent a future missbuilt.

also set a correct, SRU style version number for jammy builds going forward.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+ubuntu-core-initramfs (66.1) UNRELEASED; urgency=medium
+
+  * Add build-conflicts on snapd 2.59.3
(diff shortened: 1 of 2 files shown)


debian/changelog
```diff
@@ -1,3 +1,10 @@
+ubuntu-core-initramfs (66.1) UNRELEASED; urgency=medium
```
alfonsosanchezbeato, line 1: Won't rev 66.1 cause trouble when uploading? 67 was reverted, but even in that case.

alfonsosanchezbeato (approved):

PR #158, opened 2023-05-10
Add build-conflicts on snapd 2.59.3

xnox:
add build-conflicts on the snap-bootstrap version that has caused a recent regression. This is a safety precaution to prevent a future missbuilt.


xnox:
The CI fail is correct

```
The following packages have unmet dependencies:
 builddeps:./ : Conflicts: snapd (= 2.59.3+20.04) but 2.59.3+20.04 is to be installed
E: Unable to correct problems, you have held broken packages.
```
As indeed currently published snapd in the image ppa is the one we should not build against.

alfonsosanchezbeato (approved):
LGTM

PR #159, opened 2023-05-10
core20: fix hook permissions

xnox:
cherry-pick fix from main to ease building & signing kernel.efi snaps


alfonsosanchezbeato (approved):
LGTM

PR #162, opened 2023-05-11

alfonsosanchezbeato:


xnox:
although core22 is where we want this, can we please target this at `main` branch first? as it will make it quite easy to then combine ubuntu-core-initramfs with the fast moving snapd-edge ppa for example, to rapidly build devel series snaps with cutting edge initramfs & snapd?

@kubiko what is your feeling about it? tl;dr core-initrd just staging whichever system installed snapd is in place

alfonsosanchezbeato:
@xnox https://github.com/snapcore/core-initrd/pull/167 porposed now

PR #167, opened 2023-05-16

alfonsosanchezbeato:

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -294,10 +294,8 @@ override_dh_auto_install:
 		/usr/share/plymouth/themes/spinner/*				\
 		/usr/bin/unsquashfs						\
 		/sbin/dmsetup							\
-		/usr/lib/snapd/snap-bootstrap					\
-		/usr/lib/snapd/info						\
 		/lib/systemd/systemd-bootchart					\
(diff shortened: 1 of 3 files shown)


debian/rules
```diff
@@ -294,22 +294,15 @@ override_dh_auto_install:
 		/usr/share/plymouth/themes/spinner/*				\
 		/usr/bin/unsquashfs						\
 		/sbin/dmsetup							\
-		/usr/lib/snapd/snap-bootstrap					\
-		/usr/lib/snapd/info						\
 		/lib/systemd/systemd-bootchart					\
-		/lib/systemd/system/snapd.recovery-chooser-trigger.service;	\
+		; \
```
valentindavid, line 298: I think you use emacs. Maybe you can do `C-c C-\` which does `makefile-backslash-region`

alfonsosanchezbeato: Good one, I'll use that in the next one

xnox (approved):
Going forward it means we don't need snapd uploads into snappy-dev/image ppa, and they can be done as embargoed security uploads, SRUs into Ubuntu Archive proposed, and copied into kernel team's build-ppas as needed. Or like pulled in from edge snapd ppa, or where ever else needed from.

PR #168, opened 2023-06-06

alfonsosanchezbeato:

Backport of https://github.com/snapcore/core-initrd/pull/167

### DIFF
--- a/bin/ubuntu-core-initramfs
+++ b/bin/ubuntu-core-initramfs
@@ -273,6 +273,7 @@ def add_modules_from_file(dest_d, kernel_root, modules_d, fw_d, conf_file, db,
 
 
 def create_initrd(parser, args):
+    rootfs = "/"
     if not args.kerneldir:
         args.kerneldir = "/lib/modules/%s" % args.kernelver
     if args.root:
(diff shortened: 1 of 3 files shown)


alfonsosanchezbeato:
The kernel plugin will need https://github.com/snapcore/snapcraft/pull/4200 for things to work as usual.

alfonsosanchezbeato:
This should wait on https://github.com/snapcore/snapcraft/pull/4200 to land in snapcraft to avoid build using the plugin to fail.

xnox (commented):
I'm not sure if this will break building kernel-snaps as they expect to vendor the info file as the top level, and with this change it might not be exposed anymore. Because we copy

`cp /usr/lib/ubuntu-core-initramfs/main/usr/lib/snapd/info $(signingv)/snapd-info`

so before this lands i should update all the kernel building things to find and use alternative info file location if it doesn't exist.

Or like attempt to extract it from the initrd.

xnox:
marking this as draft for now.

alfonsosanchezbeato:
@xnox yes, I'd like to merge, I've rebased now and marked ready for review.

xnox (approved):

PR #169, opened 2023-06-12
Install in initrd

valentindavid:


alfonsosanchezbeato (commented):
Just a nitpick - also a question, how much is the initramfs increased with these changes?

Also, maybe expand a little bit the commit message and mention the binaries that are now added to the image, which seem to be:
- cryptsetup
- sfdisk
- partx
- tar
- mkfs.ext4

factory/usr/lib/the-modeenv
```diff
@@ -5,10 +5,19 @@ cp /sysroot/run/image.fstab /run/
 # TODO:UC20: move this logic into snap-bootstrap to be incorporated with 
 # initramfs-mounts
 
+if grep -q snapd_recovery_mode=run /proc/cmdline; then
```
kubiko, line 8: should we not use `/run/mnt/ubuntu-data/system-data/var/lib/snapd/modeenv` is a primary source of truth?

factory/usr/lib/the-modeenv
```diff
@@ -5,10 +5,19 @@ cp /sysroot/run/image.fstab /run/
...
+    run_mode=1
+elif grep -q snapd_recovery_mode=install /proc/cmdline; then
+    if grep -q mode=run /run/mnt/ubuntu-data/system-data/var/lib/snapd/modeenv; then
```
kubiko, line 11: this condition is too restrictive.
What if `snapd_recovery_mode=factory-reset`?

factory/usr/lib/the-modeenv
```diff
@@ -5,10 +5,19 @@ cp /sysroot/run/image.fstab /run/
...
+        run_mode=1
+    fi
+fi
+
+
 # Always bind-mount all the host filesystems
 mkdir -p /run/mnt/host
 echo '/run/mnt/host /host none rbind 0 0' >> /run/image.fstab
-if grep -q snapd_recovery_mode=run /proc/cmdline; then
+if [ -n "${run_mode-}" ]; then
```
kubiko, line 20: and then using
`if [ "${run_mode}" = "run" ]; then`

kubiko (commented):
I would prefer to reverse logic and use
`/run/mnt/data/system-data/var/lib/snapd/modeen` as primary source of truth and only if not accessible to fallback to `/proc/cmdline`

this will make it also lot easier to address other modes like `factory-reset`

factory/usr/lib/the-modeenv
```diff
@@ -5,10 +5,19 @@ cp /sysroot/run/image.fstab /run/
...
 # initramfs-mounts
 
+if grep -q snapd_recovery_mode=run /proc/cmdline; then
+    run_mode=1
+elif grep -q snapd_recovery_mode=install /proc/cmdline; then
+    if grep -q mode=run /run/mnt/ubuntu-data/system-data/var/lib/snapd/modeenv; then
+        run_mode=1
+    fi
+fi
+
```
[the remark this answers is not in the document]
valentindavid: cmdline is more  complex. It could be the last argument. In 22 branch we have a shell script that properly parses and we can use that there.

I will propose something different tomorrow.

kubiko:
@alfonsosanchezbeato with `zstd` compression this added 980kB on `arm64`
My test included `libdbus...` which we are adding regardless so even less than 980kB

factory/usr/lib/the-modeenv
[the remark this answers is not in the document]
kubiko: do you mean `/usr/libexec/core/get`?
Yeah in 22 I used 
`run_mode="$(/usr/libexec/core/get-arg snapd_recovery_mode || true)"`
If we can backport that script, of course even better, then you can do
```
if [ -e "/run/mnt/data/system-data/var/lib/snapd/modeenv" ]; then
  run_mode="$(sed -n 's/mode=\([^[:space:]]*\)/\1/p' /run/mnt/data/system-data/var/lib/snapd/modeenv)"
else
  run_mode="$(/usr/libexec/core/get-arg snapd_recovery_mode || true)"
fi
```

mvo5:
Fwiw, this looks fine to me

factory/usr/lib/the-modeenv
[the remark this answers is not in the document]
valentindavid: I have changed it. Take a look.

valentindavid:
> Also, maybe expand a little bit the commit message and mention the binaries that are now added to the image, which seem to be:

I have split it into 2 commits. So now the patch show what is added.

valentindavid:
I had to add `kmod-nls-cp437` in the modules to be able to run fsck on vfat. Not sure why it worked before without it.

debian/control
```diff
@@ -60,7 +60,7 @@ Build-Depends: debhelper-compat (= 12), dh-python, python3:any, dracut-core, qui
                squashfs-tools,
                snapd (>= 2.50+20.04),
                systemd-bootchart,
-               golang-go, indent, libapparmor-dev, libcap-dev, libfuse-dev, libglib2.0-dev, liblzma-dev, liblzo2-dev, libseccomp-dev, libudev-dev, openssh-client, pkg-config, python3, python3-docutils, python3-markdown, squashfs-tools, tzdata, udev, xfslibs-dev
+               golang-go, indent, libapparmor-dev, libcap-dev, libfuse-dev, libglib2.0-dev, liblzma-dev, liblzo2-dev, libseccomp-dev, libudev-dev, openssh-client, pkg-config, python3, python3-docutils, python3-markdown, squashfs-tools, tzdata, udev, xfslibs-dev, cryptsetup-bin
```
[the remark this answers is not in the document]
valentindavid: Done.

factory/usr/lib/the-modeenv
```diff
@@ -5,10 +5,12 @@ cp /sysroot/run/image.fstab /run/
 # TODO:UC20: move this logic into snap-bootstrap to be incorporated with 
 # initramfs-mounts
 
+mode="$(/usr/libexec/core/get-mode mode)" || mode="$(/usr/libexec/core/get-arg snapd_recovery_mode)" || mode="unknown"
```
kubiko, line 8: I like this.
I cannot find `/usr/libexec/core/get-mode` though. Have you forgot to include it in the commit?

kubiko (commented):
Looks good, just missing `/usr/libexec/core/get-mode` file to review.

kubiko (approved):
LGTM
thank you!

factory/usr/lib/modules-load.d/ubuntu-core-initramfs.conf
```diff
@@ -8,6 +8,7 @@ ahci
 libahci
 usb-storage
 nls_iso8859-1
+kmod-nls-cp437
```
alfonsosanchezbeato, line 11: Why do we need this now? Maybe it should have its own commit.

alfonsosanchezbeato (approved):
Looks very clean, thanks! Approving, but please check question.

pedronis (approved):
thanks

PR #172, opened 2023-06-19
Fix build

valentindavid:
Tar was missing from build dependencies and mkfs.ext4 was in the wrong path for focal.


alfonsosanchezbeato (approved):
Thanks!

PR #176, opened 2023-06-20
Add mkfs.vfat for u-boot gadgets

valentindavid:
U-boot needs to save `boot.sel` back to the boot partition. Because ext4 is not safe to write on some U-Boot builds due to journal corruption, we need to allow the gadget to create the boot partition as a vfat. Which means `mkfs.vfat` is required during installation.


alfonsosanchezbeato (approved):
LGTM

PR #179, opened 2023-06-22
Remove journald-console generator

valentindavid:
snapd_recovery_mode kernel parameter is not enough to know whether we need to forward messages to the console. To be able to know we need the modeenv. Journald has to be started way before that. Journald configuration cannot be reloaded. And we should not restart journald.

Users can still add `systemd.journald.forward_to_console=1` the kernel command line if they need to debug the install.


kubiko (approved):
LGTM

PR #182, opened 2023-06-27
debian: pull in systemd binaries instead of building it from scratch

alfonsosanchezbeato:
All systemd patches that were needed are now merged, so we can use the standard distro systemd.


debian/rules
```diff
@@ -276,46 +64,57 @@ override_dh_auto_install:
...
+	# But we don't won't to automount the rootfs
+	rm debian/tmp/usr/lib/systemd/system-generators/systemd-gpt-auto-generator
+	# or this (binfmt_misc module is not in the initramfs)
+	rm debian/tmp/usr/lib/systemd/system/sysinit.target.wants/proc-sys-fs-binfmt_misc.automount
+
 	# This hack should be removed with #113
 	sed -i '/^After=/{;s, *plymouth-start[.]service *, ,;/^After= *$$/d;}' debian/tmp/usr/lib/systemd/system/systemd-ask-password-*
 
-	LD_LIBRARY_PATH=`pwd`/debian/tmp/usr/lib/*/:`pwd`/debian/tmp/usr/lib/systemd `pwd`/debian/tmp/usr/bin/systemd-hwdb --root debian/tmp update --usr --strict
+	# XXX Is this still needed?
```
valentindavid, line 90: I do not find the reason in the history. Maybe the best would be to remove it and try. We can also do that in a follow up PR to make it easy to revert.

debian/rules
```diff
@@ -4,222 +4,11 @@ export DH_VERBOSE=1
...
-
-	cd vendor/systemd;					\
-	    QUILT_PATCHES=debian/patches quilt pop -a || true
-
-	rm -rf vendor/systemd/.pc
+	dh $@
 
+# bash is now used by debug-shell.service
 INSTALL_FILES_FROM_HOST=					\
+	/bin/bash						\
```
valentindavid, line 11: I think we should override it in a drop-in instead.

valentindavid (commented):
Here are some pre-review comments. I will now do some manual testing.

debian/rules
```diff
@@ -4,222 +4,11 @@ export DH_VERBOSE=1
...
-      -Dstrip=true \
-      -Dsplit-usr=false \
-      -Dsplit-bin=true \
-      -Dntp-servers=ntp.ubuntu.com \
-      -Ddns-servers='' \
-      --libdir=/usr/lib/$(DEB_HOST_MULTIARCH) \
-      -Drootlibdir=/usr/lib/$(DEB_HOST_MULTIARCH) \
-      -Dsysvinit-path='' \
-      -Dsysvrcnd-path='' \
-      -Dsulogin-path='/usr/bin/sh' \
```
valentindavid, line 108: That is also something to test. I expect that emergency shell is not usable because there is no root password.

valentindavid: We could add `Environment=SYSTEMD_SULOGIN_FORCE=1` in a drop-in maybe.

alfonsosanchezbeato: emergency mode actually worked just fine when I added bash.

valentindavid: I am missing something here on why it should work. Mantic's systemd configuration has `-Dsulogin-path=/sbin/sulogin`. And I do not even see where we copy `sulogin` into the initrd. Or create a symlink to bash.

debian/rules
xnox: systemd udev requires hwdb database, one can ship raw files of it, or just the compiled binary form (optimized), or both.
On .deb based systems we ship hwdb source files, and compile the database in maintainer scripts (such that one can optionally install systemd-hwe, or 3rd party or personal hwdb files and still have optimal boot).

Inside the initrd however, it is preffered for speed and size to compile hwdb sources files and only ship the binary database alone. See hwdb and systemd-hwdb manpages.

xnox: If it is not clear what this is doing, then please write a descriptive comment based on the above. If you drop generating a hwdb the initrd will become slower and bigger and @kubiko will haunt you.

debian/rules
xnox: do we actually need full bash? previously, and initramfs-tools initrd we have been using ash only from our busybox compilation which is smaller.

debian/rules
xnox: bash however is a huge mamonth.

are you missing busybox-initramfs in the initrd, and symlinks to all the commands it provides?

debian/rules
valentindavid: It is not about dropping hwdb. The question is why we need to move it from `/usr/lib/systemd` to `/usr/bin`?

valentindavid: I think I misread it...

debian/rules
valentindavid: I get:
```
You are in emergency mode. After logging in, type "journalctl -xb" to view
system logs, "systemctl reboot" to reboot, "systemctl default" or "exit"
to boot into default mode.
Failed to execute /sbin/sulogin: No such file or directory
```

debian/rules
```diff
@@ -276,46 +64,57 @@ override_dh_auto_install:
...
-	dh_auto_install --destdir=debian/tmp --buildsystem=meson+ninja --sourcedirectory=vendor/systemd
-	for p in $(CLEAN_PATHS); do		\
-		rm -r debian/tmp/$$p ;		\
-	done
 	touch debian/tmp/etc/machine-id
+
+	# Include all needed systemd bits
+	set -e;				\
+	for f in $$(dpkg -L systemd |	\
+	grep -E '(^/lib/systemd/system|modprobe\.d/|sysctl\.d/|rules\.d/|tmpfiles\.d/|bin/)')	\
```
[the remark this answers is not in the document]
alfonsosanchezbeato: There is no `tmp.mount` unit as far as I can see. Agreed that we should opt-in services instead of opt-out. We can do that as a follow up. I've added the `/` where needed.

debian/rules
alfonsosanchezbeato: Removed now

debian/rules
alfonsosanchezbeato: Maybe we are using different kernel command line. For me if I add
```
console=ttyS0 rd.systemd.debug_shell=ttyS0 dangerous rd.systemd.unit=emergency.service
```
I have a shell.

The interesting thing about `sulogin` is that sometimes I see the error message and sometimes I don't. Anyway, I have included it now, and I still get the shell.

We are not missing busybox bits, it is simply that `debug-shell.service` wants to run bash. I have now removed bash and dash and added an override for debug-shell to make it start sh. Said this, including bash is not a big deal compared to sizes of other things we have in the initramfs, but the more we can save the best.

valentindavid: That might have looked like working because you had `rd.systemd.debug_shell=ttyS0`.

But now it works for me. The `SYSTEMD_SULOGIN_FORCE` did not seem needed.

valentindavid: ... It works for me because of `sulogin`.

bin/ubuntu-core-initramfs
```diff
@@ -360,7 +360,6 @@ def create_efi(parser, args):
         parser.error("--stub is required, and one was not automatically detected")
     if args.root:
         args.stub = path_join_make_rel_paths(args.root, args.stub)
-        args.sbat = path_join_make_rel_paths(args.root, args.sbat)
```
xnox, line 363: this is security sensitive for our signing policy, i don't believe we can just drop sbat, as no kernel.efi will be bootable then.

why is sbat.txt support being removed? or how is it replaced?

debian/install
```diff
@@ -2,6 +2,5 @@ bin/ubuntu-core-initramfs usr/bin
 postinst.d etc/kernel/
 snakeoil/* usr/lib/ubuntu-core-initramfs/snakeoil/
 debian/tmp/* usr/lib/ubuntu-core-initramfs/main
-debian/tmp/usr/lib/systemd/boot/efi/linux*.efi.stub usr/lib/ubuntu-core-initramfs/efi/
-debian/sbat.txt usr/lib/ubuntu-core-initramfs/efi/
```
xnox, line 6: note that our copy of sbat.txt is still needed to correctly assemble kernel.efi that contains sbat section.

debian/rules
xnox: all i am hering is that to preserve previous behaviour we need to ship symlink from sulogin to sh. Given previously we configured to just call sh instead of sulogin.

debian/rules
```diff
@@ -4,224 +4,12 @@ export DH_VERBOSE=1
...
-	cd vendor/systemd;					\
-	    QUILT_PATCHES=debian/patches quilt pop -a || true
-
-	rm -rf vendor/systemd/.pc
+	dh $@
 
 INSTALL_FILES_FROM_HOST=					\
 	/bin/kmod						\
 	/bin/mount						\
+	/sbin/sulogin						\
```
xnox, line 12: previously sh was used instead of sulogin.

bin/ubuntu-core-initramfs
xnox: hm, in mantic i do see that:

# objdump -h /usr/lib/systemd/boot/efi/linuxx64.efi.stub | grep sbat
  6 .sbat         000000e0  000000000001a000  000000000001a000  00011000  2**2

sbat section is present in the stub.  maybe it's good enough.

alfonsosanchezbeato: sbat in the pre-compiled stub is fine. FTR, the current content is
```
sbat,1,SBAT Version,sbat,1,https://github.com/rhboot/shim/blob/main/SBAT.md
systemd,1,The systemd Developers,systemd,252,https://systemd.io/
systemd.ubuntu,1,Ubuntu,systemd,252.5-2ubuntu3,https://bugs.launchpad.net/ubuntu/
```
which is an updated version of what we had in the repo:
```
sbat,1,SBAT Version,sbat,1,https://github.com/rhboot/shim/blob/main/SBAT.md
systemd,1,The systemd Developers,systemd,245,https://www.freedesktop.org/wiki/Software/systemd
systemd.ubuntu,1,Ubuntu,systemd,245.4-4ubuntu3.6,https://bugs.launchpad.net/ubuntu/
```

debian/rules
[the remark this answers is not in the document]
valentindavid: Ah! There is actually a bug in the ubuntu package. It is installed in `/usr/share/systemd/tmp.mount` instead of `/usr/share/systemd/system/tmp.mount`. It has been broken for a long time. So we do not need to work-around this one for now.

debian/rules
alfonsosanchezbeato: After making `sulogin` a symbolic link to `/bin/sh` I don't get a shell if using `dangerous rd.systemd.unit=emergency.service`, it just stops execution. So I have kept `sulogin` for the moment.

alfonsosanchezbeato:
> `systemd-pcrphase` needs tpm2-tss libraries. Or needs to be removed. The same has to be done on core-base. Since we are not using PCR 11 for now, maybe we should just remove it.

I have removed these bits now.

valentindavid (approved):

xnox (approved):

PR #184, opened 2023-06-29
tests: really run FDE for fullenc spread variant

alfonsosanchezbeato:


tests/lib/nested.sh
```diff
@@ -113,15 +113,16 @@ start_nested_core_vm_unit(){
...
-    if [ "${ENABLE_OVMF_SNAKEOIL:-false}" = "true" ]; then
+        # Use test certs so our signatures can be verified
         OVMF_VARS=".snakeoil"
+        OVMF_VARS_D="/usr/lib/ubuntu-core-initramfs/snakeoil"
     fi
 
     mkdir -p "${WORK_DIR}/image/"
-    cp -f "/usr/share/OVMF/OVMF_VARS${OVMF_VARS}.fd" "${WORK_DIR}/image/OVMF_VARS${OVMF_VARS}.fd"
+    cp -f "${OVMF_VARS_D}/OVMF_VARS${OVMF_VARS}.fd" "${WORK_DIR}/image/OVMF_VARS${OVMF_VARS}.fd"
     PARAM_BIOS="-drive file=/usr/share/OVMF/OVMF_CODE${OVMF_CODE}.fd,if=pflash,format=raw,unit=0,readonly=on -drive file=${WORK_DIR}/image/OVMF_VARS${OVMF_VARS}.fd,if=pflash,format=raw"
```
valentindavid, line 126: I think we should open an issue.  I have never used this `snakeoil/OVMF_VARS.snakeoil.fd`. Never realized it was there. I doubt it will work forever. At some point it will be incompatible with `/usr/share/OVMF/OVMF_CODE.secboot.fd`.

valentindavid (approved):
I think we have to either remove `snakeoil/OVMF_VARS.snakeoil.fd` or copy `OVMF_CODE.secboot.fd`. But it will do for now.

tests/lib/nested.sh
valentindavid: I have opened https://github.com/snapcore/core-initrd/issues/187

alfonsosanchezbeato: Thanks for opening the issue! Note that this is used also for snapd tests (by downloading it from some gce url, not using it directly from this repo), and I agree that at some time we will need to upate it or build it dynamically.

PR #188, opened 2023-07-07
Trigger bootchart if ubuntu_core.bootchart is set

alfonsosanchezbeato:
in the kernel command line, instead of core.bootchart.


valentindavid (approved):
Fine to "break" a debug command line argument for core24.

If backported, I think we will need to support backward compatibility.

PR #189, opened 2023-07-07
core22: add FIPS feature

xnox:
Note this is experimental and relies on unreleased and unmerged features in the development versions of FIPS kernels.


debian/postinst
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+set -e
+
+/usr/lib/dracut/dracut-install --all --fips --ldd -D /usr/lib/ubuntu-core-initramfs/fips /usr/bin/sha512hmac || true
```
xnox, line 5: @alfonsosanchezbeato @valentindavid is this how we envision gathering skeleton at runtime, from packages that ubuntu-core-initramfs depends on? Or do you see this being listed more and more inside ubuntu-core-initramfs script?

alfonsosanchezbeato: We are already taking snapd bits when `ubuntu-core-initramfs` is run, so to be consistent this should be done from there as well. Imo doing things from there makes things more dynamic for developers building the initramfs from a container and avoids errors if you forget to reinstall the package. Not sure if there is a particular advantage when doing this from maintainer scripts.

fips/usr/lib/systemd/system/snap-initramfs-mounts.service.d/fips.conf
```diff
@@ -0,0 +1,10 @@
+[Unit]
+After=systemd-modules-load.service
+
+[Service]
+# Ensure tcrypt was loaded
+ExecStartPre=modprobe tcrypt
+# Then allow snap-bootstrap to mount things
+# Self-test verify FIPS computed hmac
+ExecStartPost=sha512hmac --check /run/mnt/kernel/.kernel.efi.hmac
```
alfonsosanchezbeato, line 9: Where does sha512hmac come from? I installed `kcapi-tools` but it does not look to be in it.

bin/ubuntu-core-initramfs
```diff
@@ -307,6 +307,17 @@ def create_initrd(parser, args):
...
 
+        if "fips" in args.features:
+            subprocess.check_call(
+                [
+                    "/usr/lib/dracut/dracut-install",
+                    "-D", main,
+                    "--ldd", "--all",
+                    "/usr/bin/sha512hmac",
+                    "/usr/bin/.sha512hmac.hmac",
+                ] + glob.glob("/usr/lib/*/.libkcapi.so.*.hmac")
```
alfonsosanchezbeato, line 318: I also cannot find this files in libkcapi1. Are these found only for package versions for fips maybe?

alfonsosanchezbeato (commented):
A PR for main would be needed as well

fips/usr/lib/systemd/system/snap-initramfs-mounts.service.d/fips.conf
xnox: Correct, only fips compliant version of kcapi-tools ships that binary in Focal & Jammy today. And this initrd must only be build against fips compliant versions of kernels & kcapi-tools. (I.e. enabling this feature alone does not make ones' kerne.efi fips compliant, one actually needs linux-fips kernel).

For mantic+ releases kcapi-hasher is shipped and will be used instead.

bin/ubuntu-core-initramfs
xnox: Correct, only fips compliant version of kcapi-tools ships hmac files. In the same place as the fips compliant kernels.

fips/usr/lib/systemd/system/snap-initramfs-mounts.service.d/fips.conf
```diff
@@ -0,0 +1,10 @@
+[Unit]
+After=systemd-modules-load.service
+
+[Service]
+# Ensure tcrypt was loaded
+ExecStartPre=modprobe tcrypt
```
alfonsosanchezbeato, line 6: But you won't need this anymore now if tcrypt is in fips.conf, right?

xnox: I do to enforce errors. modules-load.d can contain unkown modules or those that fail to load.

Whereas under FIPS compliance prior to using cryptographic functionality selftest must complete. This modprobe ensure that this is enfoced, and if tcrypt fails to load (or any selftests failed) snap-initramfs-mounts service aborts prior to mounting anything and potentially using cryptographic functionality via dm-crypt.

alfonsosanchezbeato (commented):
LGTM, thanks

alfonsosanchezbeato (approved):

PR #192, opened 2023-07-11
main: add FIPS feature

xnox:
This is a forward port of https://github.com/snapcore/core-initrd/pull/189 to mantic, with new kcapi-hasher checker, which is available in the regular Ubuntu Archive already.

Note this may change in the future, as different modules are certified, and/or if self-tests change. Note that we only build FIPS kernels for LTS releases.


xnox:
@alfonsosanchezbeato please approve this, such that i can merge it, although this will not be used until like 2025, i don't want main to lack this feature.

alfonsosanchezbeato (approved):
LGTM, thanks

PR #193, opened 2023-07-14
Switch to core24

valentindavid:


Meulengracht (approved):
Thanks, at some point we should probably establish a more formal way of how we do these things with new releases.

PR #197, opened 2023-07-19
Remove some files that are not really needed

alfonsosanchezbeato:


valentindavid (approved):
Thank you.

xnox:
I apologize for this mess. I think, these got committed by accident by me when I was still doing original shell script implementation. And I actually had them in my local ignore, so I never realized I had them in.

PR #198, opened 2023-07-24
FIPS: also support HMAC checking on FIT images

alexclewontin:
@xnox initially wrote these units to check for a hmac that is named .kernel.efi.hmac. This is correct for all systems that boot EFI UKIs. However, potentially some arm64 systems might not use EFI UKIs, but rather FIT images which we standardly name kernel.img (hence HMAC would be .kernel.img.hmac)

This (potentially naively) assumes the kernel image on any non-EFI system is named kernel.img. I think this is effectively true at the moment, especially for the subset of things we actually validate kernels for. 

This probably also should be merged into the core22 branch. It cannot go to core20 though, as ConditionFirmware was introduced into systemd after 20.04

### DIFF
--- a/fips/usr/lib/systemd/system/snap-initramfs-mounts.service.d/fips-efi.conf
+++ b/fips/usr/lib/systemd/system/snap-initramfs-mounts.service.d/fips-efi.conf
@@ -1,7 +1,11 @@
 [Unit]
 After=systemd-modules-load.service
 
+# This is for kernels that use an EFI UKI
+ConditionFirmware=uefi
+
 [Service]
+
 # Ensure tcrypt was loaded
 ExecStartPre=modprobe tcrypt
 # Then allow snap-bootstrap to mount things
--- a/fips/usr/lib/systemd/system/snap-initramfs-mounts.service.d/fips-img.conf
+++ b/fips/usr/lib/systemd/system/snap-initramfs-mounts.service.d/fips-img.conf
@@ -1,0 +1,15 @@
+[Unit]
+After=systemd-modules-load.service
+
+# This is for kernels that are not booted from an EFI environment. This probably
+# means a FIT image on arm64.
+ConditionFirmware=!uefi
+
+[Service]
+
+# Ensure tcrypt was loaded
+ExecStartPre=modprobe tcrypt
+# Then allow snap-bootstrap to mount things
+# Self-test verify FIPS computed hmac
+ExecStartPost=kcapi-hasher -n sha512hmac --check /run/mnt/kernel/.kernel.img.hmac
+# On failure, emergency.target is triggered with 10 second delay


xnox:
Reject / needs rework.

We currently do not FIPS certify things that cannot do UEFI boot.

This implementation is buggy, but I cannot comment inline from phone on how to explain it.

Yes, it is currently UEFI specific and guardrails for it should be added.

alexclewontin:
> We currently do not FIPS certify things that cannot do UEFI boot.

Is this a technical restriction, or just what we happen to do? If this is technical restriction, we need to sync offline. If it's just what we happen to do so far, this was intended as preparatory work. 

> This implementation is buggy, but I cannot comment inline from phone on how to explain it.

I'll await the explanation :)

fips/usr/lib/systemd/system/snap-initramfs-mounts.service.d/fips-img.conf
```diff
@@ -0,0 +1,15 @@
+[Unit]
+After=systemd-modules-load.service
+
+# This is for kernels that are not booted from an EFI environment. This probably
+# means a FIT image on arm64.
+ConditionFirmware=!uefi
```
xnox, line 6: Note that drops-ins (snippet in $UNIT.d/*.conf) apply for the $UNIT in question in full, not just to the unit itself.

```
$ systemctl cat hello
# /etc/systemd/system/hello.service
[Service]
Type=oneshot
ExecStart=echo hello

# /etc/systemd/system/hello.service.d/nonuefi.conf
[Unit]
ConditionFirmware=!uefi

[Service]
ExecStartPre=echo non-uefi


# /etc/systemd/system/hello.service.d/uefi.conf
[Unit]
ConditionFirmware=uefi

[Service]
ExecStartPre=echo uefi
```

Given the two drop-ins declare mutually exclusive conditions:

```
$ sudo systemctl status hello
○ hello.service
     Loaded: loaded (/etc/systemd/system/hello.service; static)
    Drop-In: /etc/systemd/system/hello.service.d
             └─nonuefi.conf, uefi.conf
     Active: inactive (dead)

Jul 24 22:24:35 xnox-Inspiron-7400 systemd[1]: Condition check resulted in hello.service being skipped.
```

It would result in snap-initramfs-mounts.service to never start, because condition failed, and will result in all systems to panic.

This seems to imply, that implementation was proposed, without design nor testing.

On this basis this is a reject, but also please learn how drop-ins works - because if such pattern is used elsewhere, it likely does not result in the behavior you desire.

fips/usr/lib/systemd/system/snap-initramfs-mounts.service.d/fips-efi.conf
```diff
@@ -1,7 +1,11 @@
 [Unit]
 After=systemd-modules-load.service
 
+# This is for kernels that use an EFI UKI
+ConditionFirmware=uefi
```
xnox, line 5: Note mutually exclusive condition in the next snippet.

fips/usr/lib/systemd/system/snap-initramfs-mounts.service.d/fips-img.conf
```diff
@@ -0,0 +1,15 @@
...
+
+[Service]
+
+# Ensure tcrypt was loaded
+ExecStartPre=modprobe tcrypt
+# Then allow snap-bootstrap to mount things
+# Self-test verify FIPS computed hmac
+ExecStartPost=kcapi-hasher -n sha512hmac --check /run/mnt/kernel/.kernel.img.hmac
```
xnox, line 14: Note, whilst it is a convention to name hmac checksum file as .$ORIG_FILE.hmac, we don't have to, as it itself encodes the filename to check.

Thus if we ever desire to support alternative filenames, we could conceivably change the .hmac file name to .kernel.hmac, encode name with extension inside it, and have a universal check, if we still use or desire such a check in the future. For example, it is likely that UEFI boot will mvoe away from this check in favour of kexec checks.

xnox (changes requested):
If this is truely required, and you do have a fips build of a reference platform that you can check fips behavour on, please provide sample builds of both kernel and kernel snap to verify the implementation and desired behavour.

For now, please open a issue in this repository with a future feature request. Please note that yes this is currently intentionally targeting UEFI platforms only.

PR #199, opened 2023-07-31
main: On arm64, attempt to decompress vmlinuz automatically.

xnox:
this will simplify a lot of things packaging side.

But also might switch to zimg soon anyway.


xnox:
Tested this against mantic on v6.5 builds the kernel inclusion is now of correct size and format out of the box, without need to manually ensure to ungzip the kernel first.

alfonsosanchezbeato (approved):
LGTM

PR #201, opened 2023-08-09
factory: create mount units specific to classic

alfonsosanchezbeato:
On mantic, systemd has changed behavior and it does not run daemon-reload from the initrd-parse-etc service if there is nothing that specifically needs that in /etc/fstab, so we were not running kernel-snap-generator again at this point, which meant we did not mount /lib/{modules,firmware} on classic systems. To fix that, force re-running the generators from a service that runs after we have the needed mountpoints.

With this kernel modules can be loaded again and snapd.seeded can progress on classic systems.

### DIFF
--- a/factory/usr/lib/systemd/system/classic-mounts.service
+++ b/factory/usr/lib/systemd/system/classic-mounts.service
@@ -7,10 +7,10 @@ DefaultDependencies=no
 
 Requires=initrd-root-fs.target
 After=initrd-root-fs.target
+Before=initrd-fs.target
 
 [Service]
 Type=oneshot
 # We force re-running kernel-snap-generator
 ExecStart=systemctl daemon-reload
-StandardOutput=journal+console
 StandardError=journal+console
--- a/factory/usr/lib/systemd/system/classic-mounts.service
+++ b/factory/usr/lib/systemd/system/classic-mounts.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Create mount units specific to classic
+OnFailure=emergency.target
+OnFailureJobMode=replace-irreversibly
+
+DefaultDependencies=no
+
+Requires=initrd-root-fs.target
+After=initrd-root-fs.target
+
+[Service]
+Type=oneshot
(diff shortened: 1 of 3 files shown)


alfonsosanchezbeato:
Maybe we should instead not use the generator, but opening for comments. Although in the end what we really want is to have this generator in a deb.

factory/usr/lib/systemd/system/classic-mounts.service
```diff
@@ -0,0 +1,16 @@
...
+DefaultDependencies=no
+
+Requires=initrd-root-fs.target
+After=initrd-root-fs.target
+
+[Service]
+Type=oneshot
+# We force re-running kernel-snap-generator
+ExecStart=systemctl daemon-reload
+StandardOutput=journal+console
```
valentindavid, line 15: Do we need console output for that?

factory/usr/lib/systemd/system/classic-mounts.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Create mount units specific to classic
+OnFailure=emergency.target
+OnFailureJobMode=replace-irreversibly
+
+DefaultDependencies=no
+
+Requires=initrd-root-fs.target
+After=initrd-root-fs.target
```
valentindavid, line 9: I think it needs `Before=initrd-fs.target`. If somehow `initrd-fs.target` is reached before the daemon-reload, I am not totally sure the generated mounts will be started.

valentindavid (commented):
Just 2 comments.

factory/usr/lib/systemd/system/classic-mounts.service
alfonsosanchezbeato: Good point, I've added this

valentindavid (approved):
Thank you.

alfonsosanchezbeato:
Merging, the issue with tests is happening due to what seems to be a problem with 23.10 image and we want this in asap.

PR #206, opened 2023-09-14
core22: On arm64, attempt to decompress vmlinuz automatically.

xnox:
This removes requirement to gzip decompress kernels before creating kernel.efi on arm.


alfonsosanchezbeato (approved):
LGTM

PR #208, opened 2023-09-15
modules/main: add support for SAS controllers storage

alfonsosanchezbeato:


xnox:
All but ses.ko are in main packages. Separately kernel team should consider promoting these out of modules extra.

xnox (approved):

xnox (approved):

PR #213, opened 2023-10-10
factory: handle properly snapd_recovery_mode based on modeenv value

kubiko:
The real source of true for the current mode is stored in `/var/lib/snapd/modeenv `this value might be altered after
`/usr/lib/systemd/system-generators/journald-console` was already run.
Or `/var/lib/snapd/modeenv` did not even exist at that time.
This happens in `install`/`factory-reset`/`recovery` modes which might be handled within the initrd.
If this is detected, update `journald-console` generator and remove already created `journald.conf.d/core-override.conf` config overlay.

### DIFF
--- a/factory/usr/lib/systemd/system-generators/journald-console
+++ b/factory/usr/lib/systemd/system-generators/journald-console
@@ -2,7 +2,11 @@
 
 set -eu
 
-if [ "$(/usr/libexec/core/get-arg snapd_recovery_mode || true)" != install ]; then
+mode="$(/usr/libexec/core/get-mode mode)" || mode="$(/usr/libexec/core/get-arg snapd_recovery_mode)" || mode="run"
+
+if [ "${mode}" = "run" ]; then
+    # remove any previous overrides
+    rm -rf /run/systemd/journald.conf.d/core-override.conf
     exit 0
 fi
 


factory/usr/lib/systemd/system-generators/journald-console
```diff
@@ -2,7 +2,7 @@
 
 set -eu
 
-if [ "$(/usr/libexec/core/get-arg snapd_recovery_mode || true)" != install ]; then
+if [ "$(/usr/libexec/core/get-arg snapd_recovery_mode || true)" == run ]; then
```
xnox, line 5: can we change this check to the same check that the-modeenv does, as in `
mode="$(/usr/libexec/core/get-mode mode)" || mode="$(/usr/libexec/core/get-arg snapd_recovery_mode)" || mode="unknown"` ?

which i guess gets the right mode both at initrd startup; and also at mid-flight reload?

kubiko:
Redone as per @xnox suggestion

xnox (approved):

alfonsosanchezbeato (approved):
LGTM, thanks!

alfonsosanchezbeato (commented):
But wait a minute... shouldn't this be actually removed by a new generator in the base? If we install from initramfs, after jumping to rootfs this generator won't be run anymore. And afaiu we want to avoid forwarding journal to console once that happens.

alfonsosanchezbeato:
We would need a PR for main

kubiko:
@alfonsosanchezbeato I have tested this on the system with compressed install steps and it does behave there as expected. So generator is re-run before root pivot

alfonsosanchezbeato:
> @alfonsosanchezbeato I have tested this on the system with compressed install steps and it does behave there as expected. So generator is re-run before root pivot

Right, because at some point there is some daemon-reload happening iirc. But that is a bit of an accident I'd say, and even more, that means that journal will stop to be sent to console while we have not switched root yet. Maybe the file could be removed in a `ExecStopPost=` stanza inside itself. A bit strange as the file would be removing itself, but I think that is fine as that information is kept in memory I think.

alfonsosanchezbeato (commented):
See comment

xnox:
@kubiko it is confusing to submit multiple PRs of the same to all series whilst review comments then spread across all of them.

Also i agree with review comments here.

Please land changes in main branch first, ahead of backporting. The code proposed here is now different to what has landed in main branch. Please fix main branch first.

PR #217, opened 2023-10-20
Allow booting UKI on systems without GPT

xnox:
This is a cherrypick from systemd 251, and thus not needed in main, but needed in UC22

This is as part of demo enablement of an EBBR like system with secureboot.


xnox:
@IsaacJT will be testing a test build of this, once it finishes building

xnox:
there are more bugs after this one, declaring bankrupt.

PR #227, opened 2024-01-03
ubuntu-core-initramfs: Fix discovery of compressed modules

valentindavid:
Now modules are compressed and have extension .ko.zstd. We need to ignore the compression extension. 

The module database was not properly generated, and was causing spurious error messages. (That database is only to find errors).


alfonsosanchezbeato (approved):
LGTM, thanks

Meulengracht (approved):
Thank you! +1

PR #228, opened 2024-01-03
ubuntu-core-initramfs: copy missing modules.builtin.modinfo

valentindavid:
This file is needed by dracut-install when installing modules. So we should copy it.

### DIFF
--- a/bin/ubuntu-core-initramfs
+++ b/bin/ubuntu-core-initramfs
@@ -340,7 +340,7 @@ def create_initrd(parser, args):
             add_modules_from_file(main, kernel_root, modules, firmware, module_load, db,
                                   warn_discoverable=True)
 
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato:
This looks fine, but I wonder how was this affecting us? Were we not including all dependencies for a module in some case?

valentindavid:
> This looks fine, but I wonder how was this affecting us? Were we not including all dependencies for a module in some case?

I suppose when a modprobe of a builtin happen it fails instead of silently ignore it. But most modprobe are done from the kernel, so they do not happen for builtin. And the others are from modules-load.d, which I think is empty now. And if not, it does not count as failure.

alfonsosanchezbeato (approved):
LGTM, thanks for the explanation

PR #229, opened 2024-01-25
factory: create components compatible mount units for kernel

alfonsosanchezbeato:
The kernel-modules components require that /lib/{modules,firmware} can be created by composition of what is shipped with the kernel plus parts that will be shipped by said component type.

For this, we need to have modinfo files in a writable directory as these files will need to be regenerated when a new kernel-modules component is installed. These files need to be readable from /lib/modules/UTS_RELEASE. Inside this directory, "kernel" and "vdso" subfolders will be bind-mounted from the kernel snap, while "updates" will have modules from the components.

In the case of UC, for modules we need an additional bind mount on top of the directory coming from the base (this is not needed for hybrid systems).

In the case of firmware, content from the kernel will be mounted to /lib/firmware as in the past. However, we will rely on the existence of a /lib/firmware/updates in the kernel to mount firware coming from components.

The mount units for the active kernel are created at boot time and stored in /run/systemd/system so they are accessible after the switch root, and are actually mounted after the switch happens, but with dependencies set so kernel modules are not loaded before the mounts are ready. This replaces the fstab/generator that was creating the mounts for modules/firmware in UC and hybrid respectively, and now unifies the approach for both.


alfonsosanchezbeato:
Creating as draft as some discussion is needed. Also, only tested on UC at the moment.

valentindavid:
It feels to me like `/usr/lib/core/kernel-mounts` could actually be a generator and moved to core-base. Everything has to be mounted after the switch root and all the information is available in at the start of systemd in main boot.

valentindavid:
> moved to core-base

(And added in the installation of snapd deb package for classic)

alfonsosanchezbeato:
> It feels to me like `/usr/lib/core/kernel-mounts` could actually be a generator and moved to core-base. Everything has to be mounted after the switch root and all the information is available in at the start of systemd in main boot.

Right, I am actually thinking about this. I thought that maybe making sure that kernel modules are available on UC auto-install would be an issue, but probably that's not the case. We might need to share the file between core-base and snapd though.

PR #230, opened 2024-02-05
ubuntu-core-initramfs: change name of plymouth plugin and move CI to noble

alfonsosanchezbeato:
Upstream has changed the name of the freetype plugin from label-ft to label-freetype. Also, we were not really running the script yet in noble, change that and perform related CI fixes.

PR #232, opened 2024-02-06
bin/ubuntu-core-initramfs: Fix plymouth freetype library name

juergh:
In Noble plymouth-label-ft, /usr/lib/<arch>/plymouth/label-ft.so was renamed to label-freetype.so.

Not sure if this needs to be backwards compatible...

### DIFF
--- a/bin/ubuntu-core-initramfs
+++ b/bin/ubuntu-core-initramfs
@@ -419,7 +419,7 @@ def install_misc(dest_dir):
         "/usr/bin/partx",
         "/usr/bin/plymouth",
         "/usr/bin/unsquashfs",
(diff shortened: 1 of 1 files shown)


kubiko:
@juergh I wonder is this is duplicate to https://github.com/snapcore/core-initrd/pull/230

juergh:
Ah yes somewhat. But there's two places where the filename needs to be fixed.

alfonsosanchezbeato:
Thanks! Closing in favor of https://github.com/snapcore/core-initrd/pull/230 , that contains also CI fixes. Thanks for pointing out that there were two places that needed to be changed!

PR #235, opened 2024-02-29
ubuntu-core-initramfs: add extra modules for nvidia-tegra-igx kernels

nwager:
Add (long) list of modules to modules/nvidia-tegra/extra-modules.conf to include necessary modules to boot UC images on nvidia-tegra devices.

Also add a check to postinst script to add nvidia-tegra feature if the kernel is nvidia-tegra-igx.

Currently, this is intended for building core22 kernel snaps with the jammy linux-nvidia-tegra-igx kernel.


nwager:
Will add custom module conf in `/usr/lib/ubuntu-core-initramfs/main/usr/lib/modules-load.d` which will allow us to make changes without changing the tool.

PR #236, opened 2024-03-05
ubuntu-core-initramfs: fix path to /sbin/modprobe

valentindavid:
The kernel will call specifically path `/sbin/modprobe`. So we need to make sure this path exists and points to modprobe.

### DIFF
--- a/bin/ubuntu-core-initramfs
+++ b/bin/ubuntu-core-initramfs
@@ -464,6 +464,8 @@ def install_misc(dest_dir):
     # /sbin/modprobe is a static path in the kernel
     os.makedirs(os.path.join(dest_dir, "sbin"))
     os.symlink("../usr/sbin/modprobe", os.path.join(dest_dir, "sbin", "modprobe"))
+    # FIXME: systemd is configured with the wrong path to dmsetup
+    os.symlink("../usr/sbin/dmsetup", os.path.join(dest_dir, "sbin", "dmsetup"))
 
 def create_initrd(parser, args):
     # TODO generate microcode instead of shipping in debian package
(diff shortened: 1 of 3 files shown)


valentindavid:
it seems we also need `/sbin/dmsetup` because systemd thinks it is there. But that sounds more like a bug in the systemd package.

valentindavid:
Also the path where to mount `/usr/lib/modules` from was prefixed with `/sysroot`, which is incorrect in our case. So we have to force it. This is a change in systemd behavior.

vorlonofportland:
Why are you proposing to create /sbin as a directory with individual symlinks here, instead of making /sbin a symlink to /usr/sbin which is consistent with what the wider ecosystem is doing wrt "usrmerge" systems?

valentindavid:
> Why are you proposing to create /sbin as a directory with individual symlinks here, instead of making /sbin a symlink to /usr/sbin which is consistent with what the wider ecosystem is doing wrt "usrmerge" systems?

To better keep track what paths need fixing. After all it is the initrd, nothing extra will run there.

vorlonofportland:
ok fair enough

Meulengracht (approved):

alfonsosanchezbeato (approved):
LGTM - the customizations for the mount units won't last long as we will not use fstab to mount `/usr/lib/{firmware,modules}` in UC24, but this is good for the moment.

alfonsosanchezbeato:
Tests seem to be failing atm though

valentindavid:
There was a typo, and it was not always failing locally. But now it should be fixed.

PR #237, opened 2024-03-13
bin/ubuntu-core-initramfs: include dbus files for org.freedesktop.systemd1

alfonsosanchezbeato:
These were pulled in on uc22, get them back as they are needed by systemd-run when used from snap-bootstrap.


andrewphelpsj (approved):
Thank you!

bin/ubuntu-core-initramfs
```diff
@@ -326,6 +326,7 @@ def install_systemd_files(dest_dir, sysroot):
     # * Configuration for systemd-tmpfiles
     # TODO: some of this can be cleaned up
     to_include = re.compile(r"^(/usr)?/lib/systemd/system|"
+                            r"/usr/share/dbus-1/.*/org.freedesktop.systemd1\..*|"
```
valentindavid, line 329: It seems we are copying `services/org.freedesktop.systemd1.service`, but this is for user session and is useless in initramfs. It is `system-services/org.freedesktop.systemd1.service` that is used. Otherwise it looks fine.

valentindavid (approved):
Looks good.

bin/ubuntu-core-initramfs
alfonsosanchezbeato: Good point, I've stopped copying that one

PR #238, opened 2024-03-14
factory: do not mount /lib/{modules,firmware} if there is a drivers tree

alfonsosanchezbeato:
Do not mount /lib/{modules,firmware} in case snapd created a drivers
tree for our kernel snap, as in that case these mounts will be created
by snapd-generator from the base or from the snapd deb package for
hybrid systems. This situation is expected to happen only for UC24+.

### DIFF
--- a/factory/usr/lib/core/extra-paths
+++ b/factory/usr/lib/core/extra-paths
@@ -2,8 +2,21 @@
 
 set -eu
 
-cat <<EOF >>/sysroot/etc/fstab
+cat <<'EOF' >>/sysroot/etc/fstab
 /run/mnt/data /writable none bind,x-initrd.mount 0 0
+EOF
+
+# Find out kernel name / revision
+kernelPath=$(cat /sys/dev/block/"$(mountpoint --fs-devno /run/mnt/kernel)"/loop/backing_file)
+kernelFile=$(basename "$kernelPath")
+kernelName=${kernelFile%%_*}
+rev=${kernelFile#*_}
+rev=${rev%%.*}
+
+# Mount /lib/{firmware,modules} only if there is no drivers tree created by snapd
+if [ ! -d /run/mnt/data/system-data/var/lib/snapd/kernel/"$kernelName"/"$rev" ]; then
+    cat <<'EOF' >>/sysroot/etc/fstab
 /run/mnt/kernel/firmware /usr/lib/firmware none bind,x-initrd.mount 0 0
 /run/mnt/kernel/modules /usr/lib/modules none bind,x-initrd.mount 0 0
 EOF
+fi
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -99,7 +99,7 @@ jobs:
           echo "************* DOWNLOADING SNAPS *************"
           download_core_initrd_snaps 'edge'
 
-          echo "************* PREPAREING CORE24 IMAGE *************"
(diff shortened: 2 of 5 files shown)


alfonsosanchezbeato:
Draft for the moment, please do not merge.

alfonsosanchezbeato:
@valentindavid @Meulengracht this is ready for review now.

valentindavid:
@alfonsosanchezbeato Could we have a reference of what PR from snapd will expect this?

valentindavid:
I suppose this was this one: https://github.com/snapcore/snapd/pull/13923

alfonsosanchezbeato:
> I suppose this was this one: [snapcore/snapd#13923](https://github.com/snapcore/snapd/pull/13923)

And https://github.com/snapcore/snapd/pull/13563 and possibly others.

factory/usr/lib/core/extra-paths
```diff
@@ -2,8 +2,14 @@
 
 set -eu
 
-cat <<EOF >>/sysroot/etc/fstab
+cat <<'EOF' >>/sysroot/etc/fstab
 /run/mnt/data /writable none bind,x-initrd.mount 0 0
+EOF
+
+# Mount /lib/{firmware,modules} only if there is no drivers tree created by snapd
+if [ ! -d /run/mnt/data/system-data/var/lib/snapd/kernel ]; then
```
valentindavid, line 10: What are the situation where this is not created? I think of 2, but I want to make sure I understand.
* When snapd is too old (will not be the case in uc26 anymore)
* It is the first boot, or factory reset

alfonsosanchezbeato: Could happen for old snapd or base < core24 so snapd does not create the drivers tree. Also on install mode or factory reset, when you have an ephemeral system. But for first boot after installation the drivers tree should be there (see https://github.com/snapcore/snapd/pull/13923).

alfonsosanchezbeato: No, we do not generate that mount in `/etc`.

In any case, as discussed it would be good to check for the path including the revision (`/var/lib/snapd/kernel/<kernel_snap_name>/<snap_revision>/`)

alfonsosanchezbeato:
@valentindavid I now check for kernel snap name / revision, I think that should handle most corner case situations. And hopefully the change guarantees that you will have a drivers tree compatible with the loaded kernel, and any core24/hybrid 24.10 system has in theory the right generator for these mounts.

I have tested the change for UC24 and hybrid in different scenarios, and seems to work as expected.

Meulengracht (approved):
Thanks, based on the changes that use the rev directly this looks good

valentindavid (approved):
Looks fine.

factory/usr/lib/core/extra-paths
alfonsosanchezbeato: Modules are mounted by the systemd generator (https://github.com/snapcore/snapd/blob/master/cmd/snapd-generator/main.c) that now we include in core24.

alfonsosanchezbeato:
The tests-snapd tests seem to fail because no eth device is created in the VM. This looks as the same as https://github.com/snapcore/core-initrd/pull/246 , so the issue does not seem directly related to the PR.

alfonsosanchezbeato:
I'd like to set `snapd.debug=1` in the kernel command line, but we need https://github.com/snapcore/snapd/pull/14077 to land first. Otherwise the tests should pass now.

PR #240, opened 2024-04-04
bin/ubuntu-core-initramfs: create manifest with used packages

alfonsosanchezbeato:
Create manifest when running "ubuntu-core-initramfs create-initrd" with debian packages and package repositories used to build the initramfs. This file will be found in path /usr/share/doc/dpkg.yaml.

### DIFF
--- a/bin/ubuntu-core-initramfs
+++ b/bin/ubuntu-core-initramfs
@@ -9,6 +9,7 @@ import glob
 import shutil
 import re
 import sys
+import hashlib
 from collections import namedtuple
 from enum import Enum, auto
 
@@ -542,10 +543,10 @@ def find_apt_repos(sysroot):
 
 
 def pkgs_used_in_initrd(dest_dir, sysroot):
-    MD5_HEX_CHARS = 32
+    MD5_HEX_CHARS = 2*hashlib.md5().digest_size
 
-    # Get list of files we have in the initramfs
-    initrd_files = []
+    # Get list of files we have in the initramfs and calculate their md5sum
+    initrd_digests = []
     for dirpath, dirs, files in os.walk(dest_dir):
         for f in files:
             path = os.path.join(dirpath, f)
@@ -554,14 +555,8 @@ def pkgs_used_in_initrd(dest_dir, sysroot):
             f_stats = os.stat(path)
             if f_stats.st_size == 0:
                 continue
-            initrd_files.append(path)
-
-    # Now calculate their md5sums
-    lines = check_output(["md5sum"] + initrd_files,
-                         cwd=dest_dir).decode("utf-8").splitlines()
-    initrd_digests = []
-    for ln in lines:
-        initrd_digests.append(ln[:MD5_HEX_CHARS])
+            dig = hashlib.file_digest(open(path, "rb"), "md5").hexdigest()
+            initrd_digests.append(dig)
 
     # Read digests from dpkg database
     root_digests = {}
--- a/bin/ubuntu-core-initramfs
+++ b/bin/ubuntu-core-initramfs
@@ -486,6 +486,139 @@ def install_misc(dest_dir, sysroot):
     # FIXME: systemd is configured with the wrong path to dmsetup
     os.symlink("../usr/sbin/dmsetup", os.path.join(dest_dir, "sbin", "dmsetup"))
 
+
+class AptRepo:
+
+    def __init__(self, url):
+        self.url = url
+        self.suites = []
+        self.components = []
+
+
+def find_apt_repos(sysroot):
+    repos = {}
+    sources = os.path.join(sysroot, "etc/apt/sources.list")
+    sources_d = os.path.join(sysroot, "etc/apt/sources.list.d")
+    state_d = os.path.join(sysroot, "var/lib/apt/lists")
+    repo_lines = check_output(["apt-cache", "policy",
+                               "-o", "Dir::Etc::SourceList="+sources,
+                               "-o", "Dir::Etc::SourceParts="+sources_d,
+                               "-o", "Dir::State::Lists="+state_d]).decode("utf-8").splitlines()
+    uri_line = re.compile(r"^ *-?[0-9]+ (http|mirror|file|cdrom|ftp|copy|rsh|ssh)")
+    repo_lines = [r for r in repo_lines if uri_line.match(r)]
+    uri_start = re.compile(r"^ *-?[0-9]+ ")
+    for r in repo_lines:
+        r = uri_start.sub("", r)
+        fields = r.split(" ")
+        if len(fields) < 2:
+            continue
+        # suite/comp in the line
+        suite_comp = fields[1].split("/")
+        if len(suite_comp) != 2:
+            continue
+        # Key will be url + suite (we only have one suite per line)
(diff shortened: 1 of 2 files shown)


bin/ubuntu-core-initramfs
```diff
@@ -486,6 +486,127 @@ def install_misc(dest_dir, sysroot):
...
+    # them point to busybox, not to the same file as in the sysroot. However,
+    # we want also the paths starting at /bin and /lib symlinks, as some
+    # packages have not done the usrmerge transition and still put files
+    # there.
+    files = check_output(["find", "./", "./bin/", "./lib/", "-type", "f"],
+                         cwd=dest_dir).decode("utf-8").splitlines()
+    # /sbin is a bit special as we are not creating that symlink, but still we
+    # get files from the debs that come from there.
+    files += check_output(["find", "./sbin/", "-type", "f"],
+                          cwd=os.path.join(dest_dir, "usr")).decode("utf-8").splitlines()
```
valentindavid, line 554: Could we use `os.walk` instead?

bin/ubuntu-core-initramfs
```diff
@@ -486,6 +486,127 @@ def install_misc(dest_dir, sysroot):
...
+
+    return final_repos
+
+
+def pkgs_used_in_initrd(dest_dir, sysroot):
+    # Get list of files in the initramfs. We do not want symlinks as some of
+    # them point to busybox, not to the same file as in the sysroot. However,
+    # we want also the paths starting at /bin and /lib symlinks, as some
+    # packages have not done the usrmerge transition and still put files
+    # there.
```
valentindavid, line 548: I see dpkg saves all the md5 of all files. If we went through all those md5 and built a hash map, then look up what package, would that solve the problems that our paths are not matching the packages?

bin/ubuntu-core-initramfs
```diff
@@ -486,6 +486,127 @@ def install_misc(dest_dir, sysroot):
...
+def find_apt_repos(sysroot):
+    repos = {}
+    sources = os.path.join(sysroot, "etc/apt/sources.list")
+    sources_d = os.path.join(sysroot, "etc/apt/sources.list.d")
+    state_d = os.path.join(sysroot, "var/lib/apt/lists")
+    repo_lines = check_output(["apt-cache", "policy",
+                               "-o", "Dir::Etc::SourceList="+sources,
+                               "-o", "Dir::Etc::SourceParts="+sources_d,
+                               "-o", "Dir::State::Lists="+state_d]).decode("utf-8").splitlines()
+    uri_line = re.compile(r"^ *-?[0-9]+ http")
```
valentindavid, line 507: Do not we have also ftp and rsync mirrors?

valentindavid (commented):
A few comments.

bin/ubuntu-core-initramfs
```diff
@@ -486,6 +486,127 @@ def install_misc(dest_dir, sysroot):
...
+    return final_repos
+
+
+def pkgs_used_in_initrd(dest_dir, sysroot):
+    # Get list of files in the initramfs. We do not want symlinks as some of
+    # them point to busybox, not to the same file as in the sysroot. However,
+    # we want also the paths starting at /bin and /lib symlinks, as some
+    # packages have not done the usrmerge transition and still put files
+    # there.
+    files = check_output(["find", "./", "./bin/", "./lib/", "-type", "f"],
```
Meulengracht, line 549: So being completely unaware of how initramfs is actually built from the packages, wouldn't it be possible to just look at folders in `/usr/share` since most packages install `copyright`s?

valentindavid: I do not think we copy those files. And if we did, we would need a way to verify that we copied the right ones. Which means we would have to look at where all the other files come from.

bin/ubuntu-core-initramfs
alfonsosanchezbeato: Good point, I've added the possible transports as specified in `sources.list(5)`

bin/ubuntu-core-initramfs
alfonsosanchezbeato: Very good suggestion, I've implemented this now. fwiw, the resulting packages are the same in the end, but using the hashes is much nicer.

bin/ubuntu-core-initramfs
alfonsosanchezbeato: In the end, `dracut_install` pulls the files from the system in the end and we do not really know the owner of them.

bin/ubuntu-core-initramfs
alfonsosanchezbeato: Not needed with the new implementation

alfonsosanchezbeato (commented):
Thanks for the reviews, this is ready now for a new pass

Meulengracht (approved):
This looks nice, I love the new approach

valentindavid (commented):
That looks better. A few more comments.

bin/ubuntu-core-initramfs
```diff
@@ -486,6 +486,139 @@ def install_misc(dest_dir, sysroot):
...
+            path = os.path.join(dirpath, f)
+            if os.path.islink(path) or not os.path.isfile(path):
+                continue
+            f_stats = os.stat(path)
+            if f_stats.st_size == 0:
+                continue
+            initrd_files.append(path)
+
+    # Now calculate their md5sums
+    lines = check_output(["md5sum"] + initrd_files,
```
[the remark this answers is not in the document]
alfonsosanchezbeato: I have used the command here as I have the impression that using hashlib would much slower. But maybe I'm wrong.

bin/ubuntu-core-initramfs
```diff
@@ -486,6 +486,139 @@ def install_misc(dest_dir, sysroot):
...
+            if sorted(repo.components) == sorted(repo_ls[j].components):
+                repo.suites += repo_ls[j].suites
+                merged.add(j)
+        final_repos.append(repo)
+
+    return final_repos
+
+
+def pkgs_used_in_initrd(dest_dir, sysroot):
+    MD5_LEN = 32
```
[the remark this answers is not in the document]
alfonsosanchezbeato: As I did not want to use hashlib, I did not want to add the dependency just because of the constant. But good point on naming, I have changed it no MD5_HEX_CHARS.

bin/ubuntu-core-initramfs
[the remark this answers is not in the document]
valentindavid: I get similar result from hashlib than md5sum.

```
[valentin@valmont ~]$ time python -c 'import hashlib; hashlib.file_digest(open("bigfile.txt", "rb"), "md5").hexdigest()'

real	0m1.442s
user	0m1.267s
sys	0m0.169s
[valentin@valmont ~]$ time md5sum bigfile.txt 
42622e2e2c67c1128cca28b030011c94  bigfile.txt

real	0m1.467s
user	0m1.337s
sys	0m0.125s
```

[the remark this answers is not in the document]
alfonsosanchezbeato: Ok, fair enough. I've changed this now. Although maybe this is not the same in all archs 😆

valentindavid (approved):
Thank you!

PR #243, opened 2024-04-10
Use ukify instead of llvm-objcopy

valentindavid:


bin/ubuntu-core-initramfs
```diff
@@ -606,35 +606,26 @@ def create_efi(parser, args):
...
-    # TODO add .osrel
     if args.cmdline:
-        cmdline = tempfile.NamedTemporaryFile(mode='w')
-        cmdline.write('%s' % args.cmdline)
-        cmdline.flush()
-        objcopy_cmd += [
-            "--add-section", ".cmdline=%s" % cmdline.name,
-            "--set-section-flags", ".cmdline=readonly,data",
-            ]
-    # TODO add .splash
```
valentindavid, line 628: I will re-add this TODO. But I want to see the tests run first.

bin/ubuntu-core-initramfs
```diff
@@ -739,35 +740,28 @@ def create_efi(parser, args):
...
+        '--os-release', '@{}'.format(args.os_release),
+    ]
+
+    if args.kernelver:
+        ukify_cmd += [
+            '--uname', args.kernelver,
+        ]
+    if args.stub:
+        ukify_cmd += [
+            '--stub', args.stub,
```
alfonsosanchezbeato, line 758: It looks like we need here `--efi-arch` too to make this arch-independent.

bin/ubuntu-core-initramfs
```diff
@@ -739,35 +740,28 @@ def create_efi(parser, args):
...
-    # TODO add .splash
-    objcopy_cmd += [
-            "--add-section", ".linux=%s" % args.kernel,
-            "--set-section-flags", ".linux=readonly,data",
-            "--add-section", ".initrd=%s" % args.initrd,
-            "--set-section-flags", ".initrd=readonly,data",
-            args.stub,
-            args.output,
+        ukify_cmd += [
+            '--cmdline', args.cmdline,
```
alfonsosanchezbeato, line 762: Will this be compatible with how we seal with FDE? I fear to expose something that might lead to misunderstandings on what is supported.

bin/ubuntu-core-initramfs
```diff
@@ -739,35 +740,28 @@ def create_efi(parser, args):
...
-    # (objdump gives a bit more of information than llvm-objdump in this case)
-    objcopy_cmd = [
-            "llvm-objcopy",
+    # TODO: add --splash
+    ukify_cmd = [
+        '/usr/lib/systemd/ukify', 'build',
+        '--linux', args.kernel,
+        '--initrd', args.initrd,
+        '--output', args.output,
+        '--os-release', '@{}'.format(args.os_release),
```
alfonsosanchezbeato, line 749: Is this simply adding a section with the content of the file or is there something else to this? What is it used for?

HACKING.md
```diff
@@ -264,28 +264,20 @@ that, to repack everything, run these commands:
...
-       tar xf - ./usr/lib/systemd/boot/efi/linuxx64.efi.stub
+$ apt install systemd-boot-efi systemd-ukify
 $ objcopy -O binary -j .linux pc-kernel/kernel.efi linux
-$ llvm-objcopy --add-section .linux=linux --set-section-flags .linux=readonly,data \
-          --add-section .initrd=initrd.img --set-section-flags .initrd=readonly,data \
-          usr/lib/systemd/boot/efi/linuxx64.efi.stub \
-          pc-kernel/kernel.efi
+$ /usr/lib/systemd/ukify build --linux linux \
+          --initrd initrd.img \
+          --output pc-kernel/kernel.efi
```
alfonsosanchezbeato, line 271: Probably worth adding the line if getting the arm64 kernel on x86, which is:
```
/usr/lib/systemd/ukify build --linux linux --initrd initrd.img --efi-arch aa64 --stub <path>/linuxaa64.efi.stub --uname <kernel_version>
```

alfonsosanchezbeato (commented):
I've tested on arm64 and this can produce a working uki, but I have some related comments, see below.

bin/ubuntu-core-initramfs
valentindavid: This was already here. I have no idea who is using this.

bin/ubuntu-core-initramfs
valentindavid: Good point.

But I wonder if was arch independent? Because I see we used `platform.machine()`.

valentindavid: Ah, I suppose if we pass `--stub

HACKING.md
valentindavid: I have added `--efi-arch aa64`. The uname should be scraped by ukify from the kernel.

But... Now I see that the stub is wrong. But it was already wrong. It did `apt download systemd-boot-efi`, which downloaded the x86-64 stub.

valentindavid: I have removed it. The section is called: `Hacking generic x86 initrd`.

alfonsosanchezbeato: `--uname` is needed if you cross-build, otherwise `ukify` produces an error:
```
# /usr/lib/systemd/ukify build --linux linux --initrd initrd.img --efi-arch aa64 --stub usr/lib/systemd/boot/efi/linuxaa64.efi.stub 
Kernel version not specified, starting autodetection 😖.
Real-Mode Kernel Header magic not found
+ readelf --notes linux
readelf: Error: Not an ELF file - it has the wrong magic bytes at the start
Found uname version: 6.8.0-20-generic
Wrote unsigned linux.unsigned.efi
```
Right, the section is about hacking x86, but probably it can be made generic or a new section can be added - anyway, that can be in a follow-up.

alfonsosanchezbeato (approved):
LGTM, thanks for the changes

alfonsosanchezbeato:
@valentindavid maybe you want to rebase the fixups before merging?

PR #244, opened 2024-04-11
tests: set ServerAliveInterval in case ssh server stops answering

valentindavid:
And also work-around cloud-init bug


alfonsosanchezbeato (approved):
LGTM, thanks

PR #248, opened 2024-07-15
modules/main/extra-modules.conf: enable dm_verity and overlay modules

sespiros:
This will allow mounting partitions with dm-verity in the initrd as well as overlayfs mounts.

### DIFF
--- a/modules/main/extra-modules.conf
+++ b/modules/main/extra-modules.conf
@@ -18,8 +18,6 @@ algif_skcipher
 cryptomgr
 dm_crypt
 dm_mod
-dm_verity
-overlay
 aes
 cbc
 echainiv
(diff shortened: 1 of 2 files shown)


modules/main/extra-modules.conf
```diff
@@ -18,6 +18,8 @@ algif_skcipher
 cryptomgr
 dm_crypt
 dm_mod
+dm_verity
+overlay
```
alfonsosanchezbeato, line 22: Could you please provide some more information on why we need this?  Is this expected to be used only for the CVM use case?

sespiros: Yes, the CVM use case is the driver for this. The change is described in spec FO164.

alfonsosanchezbeato (commented):
I think this is fine, but please add to `modules/server/extra-modules.conf` instead so we get this for non-canonical kernels.

alfonsosanchezbeato (approved):
LGTM, thanks

PR #264, opened 2024-09-18
factory: do not use overrides for /lib/{modules,firmware} mounts

alfonsosanchezbeato:
as this can override also mount units generated when we do have a drivers tree. Also, prevent crashing if the kernel has not been mounted in /run/mnt/kernel.

### DIFF
--- a/factory/usr/lib/core/extra-paths
+++ b/factory/usr/lib/core/extra-paths
@@ -6,8 +6,8 @@ cat <<'EOF' >>/sysroot/etc/fstab
 /run/mnt/data /writable none bind,x-initrd.mount 0 0
 EOF
 
-# Find out kernel name / revision
-kernelPath=$(cat /sys/dev/block/"$(mountpoint --fs-devno /run/mnt/kernel)"/loop/backing_file)
+# Find out kernel name / revision, exit if not mounted
+kernelPath=$(cat /sys/dev/block/"$(mountpoint --fs-devno /run/mnt/kernel)"/loop/backing_file) || exit 0
 kernelFile=$(basename "$kernelPath")
 kernelName=${kernelFile%%_*}
 rev=${kernelFile#*_}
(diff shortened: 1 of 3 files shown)


factory/usr/lib/core/extra-paths
```diff
@@ -6,8 +6,8 @@ cat <<'EOF' >>/sysroot/etc/fstab
 /run/mnt/data /writable none bind,x-initrd.mount 0 0
 EOF
 
-# Find out kernel name / revision
-kernelPath=$(cat /sys/dev/block/"$(mountpoint --fs-devno /run/mnt/kernel)"/loop/backing_file)
+# Find out kernel name / revision, exit if not mounted
+kernelPath=$(cat /sys/dev/block/"$(mountpoint --fs-devno /run/mnt/kernel)"/loop/backing_file) || exit 0
```
Meulengracht, line 10: why would it not be mounted?

Meulengracht (approved):
lgtm, however just a question

factory/usr/lib/core/extra-paths
```diff
@@ -19,4 +19,14 @@ if [ ! -d /run/mnt/data/system-data/var/lib/snapd/kernel/"$kernelName"/"$rev" ];
 /run/mnt/kernel/firmware /usr/lib/firmware none bind,x-initrd.mount 0 0
 /run/mnt/kernel/modules /usr/lib/modules none bind,x-initrd.mount 0 0
 EOF
+    for subdir in firmware modules; do
+        drop_d=/usr/lib/systemd/system/sysroot-usr-lib-"$subdir".mount.d
```
valentindavid, line 23: I do not think we should expect in the long term that  `/usr` will stay writable in initrd. If we start to play with sysexts then this will be read-only. I am fine with `/etc`. Though it feels like a generator thing.

alfonsosanchezbeato: Changed to `/etc`. Re: generator, that's true, but we are already touching `/etc/fstab` here for these mounts and having the definition of the drop-in files elsewhere already caused quite a bit of confusion in the past. In principle this is also a bit of a systemd bug and we should be able to remove the drop-in in the future.

valentindavid (approved):

factory/usr/lib/core/extra-paths
```diff
@@ -6,8 +6,8 @@ cat <<'EOF' >>/sysroot/etc/fstab
 /run/mnt/data /writable none bind,x-initrd.mount 0 0
 EOF
 
-# Find out kernel name / revision
-kernelPath=$(cat /sys/dev/block/"$(mountpoint --fs-devno /run/mnt/kernel)"/loop/backing_file)
+# Find out kernel name / revision, exit if not mounted
```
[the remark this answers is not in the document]
alfonsosanchezbeato: Good point, I will address in the snapd PR

alfonsosanchezbeato:
Closing in favor of https://github.com/canonical/snapd/pull/14522

PR #266, opened 2024-09-27
EFI stub: Relocate kernel image to allocate space for BSS

marcelocerri:
The kernel image needs to be relocated in order to allocate enough space for the BSS. If no space is reserved for the BSS, it can overlap with the initrd in memory, causing the corruption of it.

This change uses a similar approach as taken by upsteam in:

https://github.com/systemd/systemd/pull/33882
https://github.com/systemd/systemd/issues/33816

And fixed by systemd upstream commit id 19812661f1f6 ("stub: allocate and zero enough space in legacy x86 handover protocol").

However since the base code was vastly different, this patch basically replicates the same relocation strategy used by ARM in "vendor/systemd/src/boot/efi/linux.c".

The code could have been refactored to share code with ARM, but to minimize the risk of regressions to other architectures, linux.c was left untouched.

### DIFF
--- a/vendor/systemd-patches/0074-relocate-kernel-image-to-allocate-space-for-BSS.patch
+++ b/vendor/systemd-patches/0074-relocate-kernel-image-to-allocate-space-for-BSS.patch
@@ -1,0 +1,113 @@
+Index: systemd/src/boot/efi/linux_x86.c
+===================================================================
+--- systemd.orig/src/boot/efi/linux_x86.c	2024-09-26 15:38:46.938310743 -0300
++++ systemd/src/boot/efi/linux_x86.c	2024-09-26 17:41:15.335495467 -0300
+@@ -16,6 +16,7 @@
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato (approved):
Code LGTM - Also, I have tested this change with modified `pc-kernel` from `22` and `22-hwe`  (6.8) for x86 and things work as expected.

alfonsosanchezbeato:
Tested also on arm64 / track `22`, works fine.

chrisccoulson (approved):
This LGTM as well. I'll just copy what I wrote on Mattermost when I saw this initially. Copying the implementation used for other architectures in src/boot/efi/linux.c should be relatively safe. I know this differs slightly from the upstream code, which ensures the allocation is under 4G, but there's already nothing in this version of the stub that ensures the kernel image is within this range anyway, so I think there's a very low risk of any other regression here.

PR #271, opened 2024-10-08
bin/ubuntu-core-initramfs: warn about missing dynamic dependencies

valentindavid:
Now systemd loads most of libraries through dlopen. It also defines a .notes.dlopen section to ELF binaries to list those libraries.  We use `dlopen-notes` to list those dependencies and warn about missing ones. And eventually fail for required ones.

### DIFF
--- a/bin/ubuntu-core-initramfs
+++ b/bin/ubuntu-core-initramfs
@@ -12,6 +12,7 @@ import sys
 import hashlib
 from collections import namedtuple
 from enum import Enum, auto
+import json
 
 
 class ModTable:
@@ -398,19 +399,13 @@ def install_busybox(dest_dir, sysroot):
         os.symlink("busybox", os.path.join(dest_dir, "usr/bin", c))
 
 
-def install_misc(dest_dir, sysroot):
+def install_misc(dest_dir, sysroot, deb_arch):
     # dmsetup rules
     rules = package_files(["dmsetup"], sysroot)
     to_include = re.compile(r".*rules.d/")
     rules = [i for i in rules if to_include.match(i)]
     install_files(rules, dest_dir, sysroot)
 
-    # Other needed stuff
-    proc_env = os.environ.copy()
-    proc_env["DPKG_DATADIR"] = sysroot + "/usr/share/dpkg"
-    out = check_output(["dpkg-architecture", "-q",
-                        "DEB_HOST_MULTIARCH"], env=proc_env).decode("utf-8")
-    deb_arch = out.splitlines()[0]
     files = [
         "/usr/bin/kmod",
         "/usr/bin/mount",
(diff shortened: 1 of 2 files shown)


valentindavid:
The current missing libraries on oracular are:
```
 * libgcrypt.so.20 (suggested)
 * libtss2-esys.so.0 (suggested)
 * libtss2-rc.so.0 (suggested)
 * libtss2-mu.so.0 (suggested)
 * libidn2.so.0 (suggested)
 * libidn2.so.0 (suggested)
 * libtss2-esys.so.0 (suggested)
 * libtss2-rc.so.0 (suggested)
 * libtss2-mu.so.0 (suggested)
 * libcryptsetup.so.12 (suggested)
 * libidn2.so.0 (suggested)
 * liblzma.so.5 (suggested)
 * liblz4.so.1 (suggested)
 * libzstd.so.1 (recommended)
 * libgcrypt.so.20 (suggested)
 * libbpf.so.1 (suggested)
 * libbpf.so.0 (suggested)
 * libcryptsetup.so.12 (suggested)
 * libdw.so.1 (suggested)
 * libelf.so.1 (suggested)
 * libidn2.so.0 (suggested)
 * libarchive.so.13 (suggested)
 * libfido2.so.1 (suggested)
 * libkmod.so.2 (recommended)
 * libpwquality.so.1 (suggested)
 * libp11-kit.so.0 (suggested)
 * libqrencode.so.4 (suggested)
 * libqrencode.so.3 (suggested)
 * libtss2-esys.so.0 (suggested)
 * libtss2-rc.so.0 (suggested)
 * libtss2-mu.so.0 (suggested)
 * libip4tc.so.2 (suggested)
 * liblzma.so.5 (suggested)
 * liblz4.so.1 (suggested)
 * libzstd.so.1 (recommended)
 * libgcrypt.so.20 (suggested)
 * libidn2.so.0 (suggested)
 * libidn2.so.0 (suggested)
 * libidn2.so.0 (suggested)
 * libidn2.so.0 (suggested)
 * libidn2.so.0 (suggested)
 * libidn2.so.0 (suggested)
```

valentindavid:
Also this is useless on 24.04 as these notes do not exist yet.

bin/ubuntu-core-initramfs
```diff
@@ -616,6 +611,58 @@ def create_initrd_pkg_list(dest_dir, sysroot):
...
+                    continue
+            out = check_output(["dlopen-notes", path])
+            split = out.splitlines()
+            json_doc = b'\n'.join([s for s in split if not s[:1] == b'#'])
+            doc = json.loads(json_doc)
+            for dep in doc:
+                sonames = dep["soname"]
+                priority = dep["priority"]
+                found_sonames = []
+                for soname in sonames:
```
alfonsosanchezbeato, line 632: Is this a list of alternative implementation of the feature?

bin/ubuntu-core-initramfs
```diff
@@ -616,6 +611,58 @@ def create_initrd_pkg_list(dest_dir, sysroot):
                             pkgs).decode("utf-8")
         pkg_list.write(out)
 
+def verify_missing_dlopen(destdir, libdir):
```
alfonsosanchezbeato, line 614: This would need a comment saying what it does and why. Maybe also add an example json and maybe a link to where the format is defined. Naming the libraries that have this section would be good too (I see libsystemd-shared-256.so, but not sure if others have it).

bin/ubuntu-core-initramfs
```diff
@@ -616,6 +611,58 @@ def create_initrd_pkg_list(dest_dir, sysroot):
...
+                            continue
+                        else:
+                            missing[soname] = priority
+
+    fatal = False
+    if missing:
+        print(f"WARNING: These sonames are missing:", file=sys.stderr)
+        for m, priority in missing.items():
+            print(f" * {m} ({priority})", file=sys.stderr)
+            if priority == "required":
```
alfonsosanchezbeato, line 653: I wonder if we should require any "recommended" library. I see there libkmod.so.2 and libzstd.so.1 and both are probably useful according to "feature".

bin/ubuntu-core-initramfs
```diff
@@ -616,6 +611,58 @@ def create_initrd_pkg_list(dest_dir, sysroot):
...
+            for dep in doc:
+                sonames = dep["soname"]
+                priority = dep["priority"]
+                found_sonames = []
+                for soname in sonames:
+                    dest = os.path.join(destdir, os.path.relpath(libdir, "/"), soname)
+                    if os.path.exists(os.path.join(destdir, dest)):
+                        found_sonames.append(soname)
+                if not found_sonames:
+                    for soname in sonames:
```
alfonsosanchezbeato, line 637: This needs a comment to help the reader (afaiu this is for having the highest priority if the library is for more than one feature).

alfonsosanchezbeato (commented):
Thanks for this! I have some comments, mainly about documentation.

alfonsosanchezbeato:
Found the reference now: https://systemd.io/ELF_DLOPEN_METADATA/

bin/ubuntu-core-initramfs
valentindavid: No, it is a list of shared object names. That is it might try to dlopen `libfoo.so.1`, and if that does not work `libfoo.so.2`. In that case the sonames will be a list of 2 strings.

bin/ubuntu-core-initramfs
valentindavid: I think at this point only "required" is used. But maybe I have missed something.

alfonsosanchezbeato: I see some recommended in oracle:
```
root@oracular-confined:~# dlopen-notes -s /usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-256.so  | grep recommended
libkmod.so.2 recommended
libzstd.so.1 recommended
```

bin/ubuntu-core-initramfs
alfonsosanchezbeato: Both things are true, aren't they? :slightly_smiling_face:

bin/ubuntu-core-initramfs
valentindavid: I was wrong, "required" is never used by systemd.  I will make missing "recommended" libraries fatal.

alfonsosanchezbeato (approved):
LGTM, thanks

PR #276, opened 2024-11-29
bin/ubuntu-core-initramfs: add /sbin/reboot symlink

valentindavid:
snap-bootstrap needs to call /sbin/reboot in recover mode when remodeling.


Meulengracht (approved):
thank you for the quick fix!

alfonsosanchezbeato (approved):
LGTM, thanks

PR #279, opened 2024-12-05
factory: mount drivers tree when installing from initramfs

alfonsosanchezbeato:
In the case when we were installing from the initramfs, we were not mounting the generated drivers tree. Fix that by looking for the kernel revision in modeenv in that case.


alfonsosanchezbeato:
Depends on https://github.com/canonical/core-initrd/pull/262

alfonsosanchezbeato:
This will be performed by snap-bootstrap now, see https://github.com/canonical/snapd/pull/14810

PR #281, opened 2025-01-15
tests: log spread failures to grafana

maykathm:


alfonsosanchezbeato:
@maykathm note that core-initrd is now part of snapd for UC24+, although we will still use this repo for older releases. Although those are in specific branches.

PR #294, opened 2026-03-02
core24: d/copyright: Add correct license for fonts files

nishitm:
Update d/copyright file for core24 as font files have a different license coming from the `fonts-ubuntu` package


alfonsosanchezbeato:
This needs to be proposed to https://github.com/canonical/snapd/tree/master/core-initrd instead

